Issue #10, opened 2026-04-16
Process tree loses intermediate exec steps on same-PID re-execs


getvictor:
## Problem

When a shell uses exec-optimization (e.g., `sh -c "command"` calls `execve()` directly), the PID is reused across multiple exec calls. Our `processes` table upserts on PID, so only the final executable path is stored. This causes the process tree to skip intermediate steps.

**Example:** `python → sh → bash → /tmp/payload` all happen on PID 30683 via successive `execve()` calls. The tree shows `python → payload`, losing the `sh` and `bash` nodes entirely.

This matters for security investigations -- analysts need to see the full execution chain to understand how a payload was launched.

## How other EDRs solve this

- **Elastic Defend**: Assigns a unique `process.entity_id` (UUID) per exec event, independent of PID. Tree is built on entity IDs, so re-execs are naturally separate nodes.
- **CrowdStrike Falcon**: Records every `execve()` as a distinct event with a re-exec indicator. Same PID appears as multiple linked nodes.
- **Microsoft Defender**: Uses composite key `(PID, start_time)` -- each exec resets the start time, creating a new node.
- **Carbon Black**: Tracks "process segments" per exec on the same PID.

## Proposed fix

1. **Server processor**: Store each exec event as a new row instead of upserting on PID. Use `(pid, exec_timestamp)` or a generated UUID as the unique identity.
2. **Process tree API**: Return re-exec chains when building the tree, connecting same-PID execs in order.
3. **UI tree renderer**: Display re-exec nodes with a visual indicator (e.g., dashed edge or "re-exec" label) so analysts can distinguish a re-exec from a fork+exec.

## Context

Discovered while investigating alert #20 -- `python` spawned `sh -c "/tmp/payload"`, which exec-optimized through `sh → bash → /tmp/payload` on PID 30683. The raw `events` table has all three exec records, but the materialized `processes` row only has the final path.

getvictor:
Resolved by PR #34 (commit 26adefe 'A2 (issue #10): preserve same-PID re-exec chains'). The `processes` row identity moved off PID-only to `(host_id, pid, exec_time_ns)` with `previous_exec_id` chaining, and `graph.BuildTree` stitches re-exec branches under the same fork parent so the UI renders the chain.

PR #34, opened 2026-04-24
A2 (issue #10): preserve same-PID re-exec chains

getvictor:
Shell exec-optimization chains — python → sh → bash → /tmp/payload on a single PID via successive execve() calls — used to collapse because processes upserted on PID. Only the final path survived. Analysts lost the intermediate steps that explain HOW a payload was launched.

Server-side storage + materializer:
- processes.previous_exec_id: new nullable FK-like column linking a row back to the prior generation. First row in a chain has NULL.
- ExitReasonReExec ("reexec"): distinguishes synthesized re-exec closes from observed exits, TTL reconciliations, and PID-reuse closes.
- Builder.handleExec: three-way branch on the current-row state. No row → exec-without-fork synthesis (unchanged). Row with exec_time_ns=NULL → first exec after fork (UPDATE, unchanged). Row with exec_time_ns set → re-exec path: close prior row (exit_time_ns = new exec time, exit_reason = reexec), INSERT new row preserving fork_time_ns and PPID, link via previous_exec_id.

Store/API surface:
- CloseReExecGeneration(ctx, rowID, exitTimeNs, exitIngestedAtNs) — writes the synthesized exit for the predecessor row.
- GetExecChain(ctx, current) — walks previous_exec_id backward, bounded at 64 hops, returns oldest-first.
- ProcessDetail.ReExecChain exposed via GET /api/v1/hosts/{id}/processes/{pid}.

UI:
- types.ts: Process grows exit_reason + previous_exec_id; ProcessDetail grows re_exec_chain. Structural types stay backward-compatible.
- ProcessDetail.tsx: renders "Re-exec chain (N prior generations)" block listing each prior path + exec time + "— current" for the live generation. Hidden when the chain is empty.

Tests:
- server/graph/reexec_test.go: reproduces the canonical issue #10 scenario (fork pid=30683, then 4 execs on the same pid with no intervening fork) and asserts 4 rows, chain of 3 priors via GetDetail, preserved fork_time_ns across the chain, exit_reason ="reexec" on each prior generation. Includes a negative test that a plain fork+exec produces no chain.

Validated live:
- Local smoke server: POST the 5-event chain, verify DB has 4 rows correctly linked, JSON API returns the chain oldest-first.
- Brave visual smoke: logged in, deep-linked to the process via ?process=N&at=, observed the Re-exec chain section rendering "/usr/bin/python3 → /bin/sh → /bin/bash → /tmp/payload (current)".

VM validation deferred to Phase 7 Track E (dogfood QA): the server processing is identical for ES-sourced and HTTP-POST-sourced events, and the VM would only exercise Apple's documented NOTIFY_EXEC-per- execve behavior.

## Release Notes

**New Features**
- Process re-execution tracking: the system now tracks and displays the execution history when a process executes multiple different programs. The process detail view shows each program execution with timestamps, giving visibility into all programs executed by a single process throughout its lifetime.

### DIFF
--- a/server/graph/builder.go
+++ b/server/graph/builder.go
@@ -114,52 +114,97 @@ func (b *Builder) handleExec(ctx context.Context, evt store.Event) error {
 		return err
 	}
 
-	// Try to update an existing process record (from a prior fork).
-	err := b.store.UpdateProcessExec(ctx, store.ProcessExecUpdate{
-		HostID:      evt.HostID,
-		PID:         p.PID,
-		ExecTimeNs:  evt.TimestampNs,
-		Path:        p.Path,
-		Args:        p.Args,
-		UID:         p.UID,
-		GID:         p.GID,
-		CodeSigning: p.CodeSigning,
-		SHA256:      p.SHA256,
-	})
+	// Resolve the running-at-this-moment row for the PID. Three shapes:
+	//   (a) no row            — exec-without-fork (synthesize a root row).
+	//   (b) row, exec_time_ns NULL  — first exec after fork; UPDATE in place.
+	//   (c) row, exec_time_ns set   — same-PID re-exec (issue #10): close the
+	//       prior generation and INSERT a new linked row. Without this branch,
+	//       shell exec-optimization chains (python→sh→bash→/tmp/payload) get
+	//       collapsed into the final exec only.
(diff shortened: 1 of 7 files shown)


getvictor:
@coderabbitai full review

PR #35, opened 2026-04-24
D1: embed Redoc at /api/docs with zero external network calls

getvictor:
A self-hosted API documentation page backed by Redoc, compiled into the server binary. Pilots hit `https://<server>/api/docs` and see the full OpenAPI 3.1 spec rendered in the browser without logging in, without leaving their own network.

New package `server/api/docs`:

- `embed/redoc.standalone.js` — Redoc 2.x open-source bundle, patched so the upstream `https://cdn.redoc.ly/redoc/logo-mini.svg` reference is rewritten to `/api/docs/logo-mini.svg`. The upstream bundle makes that one external call for Redoc's brand mark in the top nav; for an EDR product where air-gapped pilots are a real use case, even a tiny third-party-CDN dependency is wrong. The logo is MIT-licensed per the Redoc repo, so the same-origin rewrite is legal.
- `embed/logo-mini.svg` — the brand mark, downloaded once and shipped with the binary.
- `embed/openapi.yaml` — the canonical spec mirrored from `docs/api/openapi.yaml`. A `go:generate` directive copies the canon file into the embed dir; run `go generate ./server/api/docs/...` after editing the canon to keep them in sync (future follow-up: CI check that flags drift).
- `docs.go` — RegisterRoutes wires four paths onto the public mux:
  GET /api/docs                     → HTML bootstrap
  GET /api/docs/redoc.standalone.js → the bundle
  GET /api/docs/logo-mini.svg       → the brand mark
  GET /api/openapi.yaml             → the spec
- `bundleHash` is a 12-char content-hash computed once at init, used
  as a ?v=... cache-buster on the script src in the served HTML. When
  Redoc is updated, the hash changes and browsers refetch; within a
  release the bundle has a 1-hour public cache, so repeat visits are
  fast without a revalidation round-trip.

Three tests cover the contract:

- TestRegisterRoutes_IndexServesHTMLReferencingAssets: the inline HTML points at the embedded spec + bundle and contains no external URLs.
- TestRedocBundle_NoExternalURLs: regression test guaranteeing no cdn.redoc.ly string ever sneaks back in (e.g. someone re-downloads a fresh Redoc bundle and forgets to re-patch).
- TestRegisterRoutes_SpecAndBundleServedFromEmbed: all three asset routes return 200 with the right Content-Type and recognizable leading bytes.

Live smoke in Brave (phase-7-d1 local build):
- Redoc renders cleanly with all 29 tag headings (Agent, Session, Hosts, Alerts, Commands, Admin, etc.) visible.
- performance.getEntriesByType('resource') shows 4 resources total, all from localhost:18089. external array is empty.
- Zero console errors on load.

docs/api.md now links to the live renderer with a note that both endpoints are public and served from the same binary.

## Summary by CodeRabbit

* **New Features**
  * Added self-hosted API documentation accessible at `/api/docs` with an interactive specification viewer
  * OpenAPI 3.1 specification now available at `/api/openapi.yaml`
  * All documentation assets served directly from the server without external dependencies

* **Documentation**
  * Updated API documentation to clarify runtime endpoints for accessing the OpenAPI specification and rendered docs


getvictor:
@coderabbitai full review

PR #38, opened 2026-04-25
B3 round 2: privilege_launchd_plist_write detection rule

getvictor:
Pairs with persistence_launchagent (T1543.001) by catching the
system-domain side of LaunchDaemon persistence (T1543.004). Fires on
write-mode open() against any *.plist directly under
/Library/LaunchDaemons/, skipping Apple platform binaries (installd,
system_installd, sysadminctl) and operator-allowlisted MDM agents.

Why catch the file write rather than the launchctl bootstrap activation:
LaunchDaemon activation is often deferred to reboot; the file drop is
the moment we want to surface to operators. The activation-side
catch is what persistence_launchagent already covers for user agents,
and adding a system-domain twin rule there is straightforward later.

* Rule + 7 fixture cases (3 positive: O_WRONLY, O_WRONLY|O_CREAT|O_TRUNC,
  unsigned-writer; 4 negative: read-only, platform binary writer,
  team-ID allowlisted, nested subdirectory).
* EDR_LAUNCHDAEMON_TEAMID_ALLOWLIST env var lets operators whitelist
  legitimate non-Apple agents (Munki, Kandji, JumpCloud) without
  silencing the rule wholesale.
* Wires the rule into main.go's engine registration and into the
  all_rules_integration_test.go aggregate; the test now asserts
  privilege_launchd_plist_write fires alongside the other six rules
  on a single event batch.
* Docs updated (operations.md + install-server.md env table) so
  pilot operators can find the allowlist knob.

Known limitations documented in the rule godoc:
- Atomic writes via temp file + rename: ESF NOTIFY_OPEN sees the
  temp path, not /Library/LaunchDaemons/<name>.plist. Tracked for
  Phase 8 (NOTIFY_RENAME subscription).
- Writes mediated by a platform binary (e.g. attacker uses sudo cp):
  cp is platform-signed, so this rule skips. The parent shell's exec
  is captured by suspicious_exec / process tree.

### DIFF
--- a/docs/install-server.md
+++ b/docs/install-server.md
@@ -206,6 +206,7 @@ unset uses the documented default.
 | `EDR_RETENTION_DAYS` | no | 30 | Event TTL, 0 disables retention |
 | `EDR_RETENTION_INTERVAL` | no | 1h | How often the retention job runs |
 | `EDR_LAUNCHAGENT_ALLOWLIST` | no | — | Comma-separated absolute paths the `persistence_launchagent` rule treats as benign |
+| `EDR_LAUNCHDAEMON_TEAMID_ALLOWLIST` | no | — | Comma-separated code-signing team IDs the `privilege_launchd_plist_write` rule treats as benign |
 | `EDR_LOG_LEVEL` | no | info | `debug` / `info` / `warn` / `error` |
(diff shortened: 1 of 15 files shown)


getvictor:
@coderabbitai full review

PR #40, opened 2026-04-25
E1: synthetic-attacker runbook for Phase-7 dogfood

getvictor:
scripts/qa/attack-runbook.sh fires every shipped detection rule
exactly once against an EDR-instrumented host. Designed to ssh into
the dev VM and run, then watch alerts roll into the admin UI.

Six positive steps (one per shipping rule), each with a comment
explaining how the synthetic input maps to the rule's positive
trigger. Notes:

* shell_from_office is intentionally not exercised — Office isn't
  on the dev VM, the rule keys on /Applications/Microsoft *.app
  parents, and the plan tracks it via the Swift test harness instead.
  Listed as INTENTIONALLY NOT exercised in the final summary so
  operators don't think the script is broken.
* privilege_launchd_plist_write requires a non-platform-binary
  writer (the rule skips Apple-signed binaries like cp / dd / tee
  / bash). The step compiles a tiny Go dropper in the working dir
  to satisfy that — the closest scriptable stand-in for "an
  attacker drops a daemon plist". Skipped with a warning when no
  Go toolchain is present.
* All steps produce no real malware. /tmp scratch only, plist
  drops are removed by the dropper itself, network calls go to
  127.0.0.1 so air-gapped runs work.
* set -uo pipefail (no -e) plus a logging ERR trap: every step
  runs to completion regardless of whether the previous one
  succeeded, so a partial failure leaves the operator with a
  complete picture of which alerts fired and which didn't.

Validated: bash -n, shellcheck clean. End-to-end run is the
Phase-7 QA milestone (E5) once the agent + server are wired on
the VM — that's a future commit pinning down the report.

## Summary by CodeRabbit

* **Tests**
  * Added a new QA runbook Bash script with sequential execution of multiple benign test actions designed to trigger specific detection rules. The script includes temporary work directory management, granular error handling with step-level logging that continues on errors, and post-execution verification guidance with expected alert descriptions and cleanup instructions for operators.

### DIFF
--- a/scripts/qa/attack-runbook.sh
+++ b/scripts/qa/attack-runbook.sh
@@ -1,0 +1,236 @@
+#!/usr/bin/env bash
+#
+# Phase-7 dogfood attack runbook (issue: claude/mvp/phase-7-pilot-hardening.md E1).
+#
+# What this is
+# ============
+# A reproducible synthetic-attacker script that fires every shipped detection
+# rule exactly once. Run on a host with the EDR agent installed + enrolled,
+# then check the EDR admin UI for one alert per step. The script is
+# explicitly NOT malicious — every step generates a benign artifact under
+# /tmp or removes itself; nothing persists, nothing exfiltrates real data.
+#
+# Why each step looks like it does
+# --------------------------------
+# Every rule maps onto one step. The shape of the step is reverse-engineered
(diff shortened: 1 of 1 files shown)


getvictor:
@coderabbitai full review

PR #41, opened 2026-04-25
B3 round 3: sudoers_tamper detection rule

getvictor:
Closes Track B for Phase 7 (rules 3-of-3). Pairs with credential_keychain_dump (T1555.001) and privilege_launchd_plist_write (T1543.004) by catching the third macOS post-exploitation primitive listed in the plan: writes to /etc/sudoers or /etc/sudoers.d/* by a non-allowlisted process. ATT&CK T1548.003 (Abuse Elevation Control Mechanism: Sudo and Sudo Caching).

Design notes worth recording for future rule authors:

* Unlike the launchd-plist rule, this one does NOT key on is_platform_binary. The canonical attacker tools for sudoers tampering are platform binaries themselves: cp, tee, /bin/bash (shell redirection), even sudo vi. Filtering platform binaries would silence every realistic attack pattern. Instead we fire on every direct write and let operators tune via EDR_SUDOERS_WRITER_ALLOWLIST.
* visudo / sudoedit are NOT in the default allowlist because they don't need to be: both write a temp file and atomically rename it onto /etc/sudoers. ESF NOTIFY_OPEN doesn't fire on rename, so the rule never sees the visudo flow at all.
* The path regex matches both /etc/sudoers and /private/etc/sudoers forms because ESF reports the path as the caller passed it (and /etc is a symlink to /private/etc on macOS). Subdirectories like /etc/sudoers.d/staging/x are excluded by the trailing [^/]+ so the rule can't be lured by an attacker creating a same-named subdirectory somewhere unrelated.

Known limitation: an attacker with root could create the temp file and rename it onto /etc/sudoers without firing NOTIFY_OPEN at all. NOTIFY_RENAME would catch that, but the extension doesn't subscribe to it today (tracked alongside the privilege_launchd_plist_write atomic-rename gap).

Implementation:
* server/detection/rules/sudoers_tamper.go — rule + fast-path bytes filter + path regex + write-mode check + writer-allowlist gate.
* 7 fixture cases (3 positive: shell redirect, cp overwrite, /private/etc/ path; 4 negative: read-only open, allowlisted writer, unrelated path /etc/sudoers.example, nested /etc/sudoers.d/staging/x).
* Go-level unit tests for allowed() edge cases, malformed-payload drop, and the proc==nil race guard. Coverage of the new file: ID/Techniques/allowed at 100%, Evaluate 87.5%, evalEvent 89.5% — the two uncovered lines are pure store-error wrappers.
* Wires through Config.SudoersWriterAllowlist (env var EDR_SUDOERS_WRITER_ALLOWLIST), main.go registration, and the all_rules_integration_test aggregate.
* Docs: operations.md + install-server.md env table updated.

The rule's openPayload struct + write-access mask are named sudoersOpenPayload / sudoersWriteAccessMask to dodge a future merge collision with PR #38's privilege_launchd_plist_write rule, which defines parallel symbols. Once both rules are on main a follow-up can dedupe to a shared open_event.go.

### DIFF
--- a/docs/install-server.md
+++ b/docs/install-server.md
@@ -207,7 +207,7 @@ unset uses the documented default.
 | `EDR_RETENTION_INTERVAL` | no | 1h | How often the retention job runs |
 | `EDR_LAUNCHAGENT_ALLOWLIST` | no | — | Comma-separated absolute paths the `persistence_launchagent` rule treats as benign |
 | `EDR_LAUNCHDAEMON_TEAMID_ALLOWLIST` | no | — | Comma-separated code-signing team IDs the `privilege_launchd_plist_write` rule treats as benign |
-| `EDR_SUDOERS_WRITER_ALLOWLIST` | no | — | Comma-separated writer-process absolute paths the `sudoers_tamper` rule treats as benign |
+| `EDR_SUDOERS_WRITER_ALLOWLIST` | no | — | Comma-separated writer-process absolute paths the `sudoers_tamper` rule treats as benign; alerts may surface either `/etc/sudoers...` or `/private/etc/sudoers...` because `/etc` is a symlink and ES reports the path as opened |
 | `EDR_LOG_LEVEL` | no | info | `debug` / `info` / `warn` / `error` |
 | `EDR_LOG_FORMAT` | no | json | `json` or `text` |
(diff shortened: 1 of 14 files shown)


getvictor:
@coderabbitai full review

Issue #11, opened 2026-04-16
Enumerate running processes at ESF startup for baseline visibility


getvictor:
## Problem

ESF (Endpoint Security Framework) is a pure event stream -- it only captures events that occur after the client subscribes. Processes that were already running before the extension starts (or restarts) have no fork/exec events in the database, so they never appear in the process tree.

**Example:** Safari (PID 33392) was running before the extension activated. Its WebKit helper XPC services (PIDs 33357-33362) were launched *after* activation and show up fine. But Safari's main process is invisible in the tree despite actively generating `open` events (proving the ESF client can see it).

This is a known ESF limitation that every production EDR must handle.

## Proposed fix

At ESF client startup, before entering the event loop, enumerate all running processes using `proc_listallpids()` and `proc_pidpath()`, then synthesize exec events for each. This gives the server a baseline snapshot of every process on the host.

Steps:
1. Call `proc_listallpids(nil, 0)` to get the count, allocate a buffer, then call again to fill it
2. For each PID, call `proc_pidpath()` to get the executable path
3. Optionally read PPID via `sysctl(CTL_KERN, KERN_PROC, KERN_PROC_PID, pid)` for parent info
4. Emit synthetic exec events (with a flag like `"snapshot": true` in the payload so the server can distinguish them from real execs)
5. Begin normal ESF event subscription

## How other EDRs handle this

- **CrowdStrike Falcon**: Snapshots all running processes at sensor startup, then merges with the live event stream
- **Elastic Defend**: Uses `proc_listallpids` on macOS for initial enumeration, marks them as "already_running" in the process tree
- **osquery**: Periodically queries the full process table via `proc_listallpids` (poll-based, not event-based)

## Impact

Without this fix, any process launched before the extension starts is a blind spot -- it won't appear in the process tree, won't be attributed in network connections, and won't trigger detection rules. This is especially problematic after extension restarts or system reboots where long-lived daemons and apps are already running.

getvictor:
Resolved by #174 (ESF startup process snapshot for baseline visibility).

PR #42, opened 2026-04-25
A4 server-side: tolerate + filter snapshot exec events (issue #11)

getvictor:
Server half of the process-baseline-enumeration work. The agent half (ESF subscriber walks proc_listallpids at startup, emits a synthetic exec per live PID) lands separately in a Swift PR; this commit prepares the data plane so those events have somewhere to land without misbehaving.

What this changes:

* schema/events.json — exec_payload grows an optional `snapshot` bool. Default false. Agents stamp `snapshot=true` on the synthetic exec events they emit during the startup baseline pass.
* server/detection/engine.go — Engine.Evaluate now filters `snapshot=true` exec events out of the per-rule batch before dispatching. The filter uses a bytes.Contains fast-path (`"snapshot":true`) plus a JSON-decode probe to keep the per-event cost negligible — ESF NOTIFY_EXEC fires often, so a free unmarshal per event isn't acceptable. Snapshot events still flow through ingest, still hit `events`, still get materialised by graph.Builder.insertExecWithoutFork into the process tree; they just don't trigger detection rules.

Why filter at the engine, not in each rule: rules describe new attacker activity. Snapshot events describe state that existed before the extension subscribed. Letting individual rules see them would generate false positives every extension restart — e.g. dyld_insert firing on a Safari debug-build with
DYLD_INSERT_LIBRARIES set, or shell_from_office firing on a long-running Word session that pre-dated the EDR install. Centralising the gate in the engine keeps rules simple and means future rules don't have to remember the check.

Tests:
* TestFilterSnapshotEvents — table-driven unit test on the helper. Covers snapshot=true, snapshot=false, missing field, non-exec event types, malformed JSON, and the false-positive guard ("\"snapshot\":true" appearing as a literal string in argv).
* TestEngineEvaluateDropsSnapshotExecs — full engine integration: registered rule sees live exec + fork events, never the snapshot exec.

Documented gotcha in the test comment: the bytes.Contains gate is byte-exact on the Go-encoded shape (`"snapshot":true` with no spaces). ESF events come from a typed Swift encoder that emits no extra whitespace, so the spaced form `"snapshot": true` doesn't appear in production. If Swift's encoder ever changes its formatting we'll need to widen the gate; the test pins the current behaviour so the regression is loud.

Build, lint, and full detection + ingest test suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection accuracy by filtering out synthetic startup events from rule evaluation, preventing false positives on extension restart.

* **Tests**
  * Added test coverage for startup event filtering and detection engine behavior.

### DIFF
--- a/schema/events.json
+++ b/schema/events.json
@@ -59,7 +59,11 @@
         "uid": { "type": "integer" },
         "gid": { "type": "integer" },
         "code_signing": { "$ref": "#/definitions/code_signing" },
-        "sha256": { "type": "string" }
+        "sha256": { "type": "string" },
+        "snapshot": {
+          "type": "boolean",
+          "description": "True for synthetic exec events emitted at extension startup to materialise processes that existed before ESF subscription (issue #11). Detection rules skip snapshot events because they describe historical state, not new attacker activity."
+        }
       }
     },
     "fork_payload": {
--- a/server/detection/engine.go
+++ b/server/detection/engine.go
@@ -1,7 +1,9 @@
 package detection
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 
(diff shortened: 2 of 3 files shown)


getvictor:
@coderabbitai full review

PR #44, opened 2026-04-25
Apply entitlements when re-signing release artifacts (Phase-7 ship bl…

getvictor:
…ocker)

v0.1.0-rc.3's host app + both system extensions ship with empty entitlements blobs, even though the per-binary entitlement plists exist in extension/edr/{edr,extension,networkextension}/*.entitlements and Xcode references them at build time.

Root cause: build.sh re-signs all three bundles after embedding the provisioning profile (which invalidates the Xcode-time signature), but the bare `codesign --sign` calls don't pass `--entitlements`, so the resulting binaries are signed with an empty entitlements blob.

User-visible symptom on a SIP-enabled (i.e. pilot-customer) Mac:
* Install the .pkg, double-click /Applications/Fleet EDR.app.
* Host app silently submits OSSystemExtensionRequest.activationRequest.
* macOS rejects the request before it reaches sysextd because the caller lacks `com.apple.developer.system-extension.install`.
* The "System Extension Blocked — Click to allow" prompt that System Settings normally shows never appears, because nothing was ever staged.
* Agent log spams `xpc error code 1, expected` forever as it retries XPC to extensions that aren't loaded.

This was masked all along by edr-dev being SIP-disabled — without SIP enforcing entitlements, ESF + NE load anyway. The bug only surfaces on the SIP-on QA VM that's meant to simulate a real pilot customer Mac, which is exactly where it matters.

Fix: each post-Xcode `codesign --sign` re-sign call now passes `--entitlements <file>` with the matching plist. The plists already exist and have the right contents:
  * edr.entitlements — com.apple.developer.system-extension.install + networkextension types
  * extension.entitlements — com.apple.developer.endpoint-security.client
  * networkextension.entitlements — networkextension provider types
    + application-groups for IPC

Also added explicit existence checks so a missing plist fails the build loudly instead of silently producing another broken release.

Validated locally with bash -n + shellcheck. The next release tag will exercise this end-to-end; the canary is "OSSystemExtensionRequest no longer fails immediately on SIP-on hosts" → System Settings shows the approval prompt as expected.

## Summary by CodeRabbit

* **Chores**
  * Improved the build process to apply per-binary entitlements during code signing for the EDR app and system extensions, with validation to ensure entitlements files are present.

### DIFF
--- a/packaging/pkg/build.sh
+++ b/packaging/pkg/build.sh
@@ -184,7 +184,17 @@ else
 
     PROFILE_NET="$ROOT/packaging/provisioning/networkextension.provisionprofile"
     NETEXT="$STAGE/app-root/Applications/Fleet EDR.app/Contents/Library/SystemExtensions/com.fleetdm.edr.networkextension.systemextension"
-    if [ -f "$PROFILE_NET" ] && [ -d "$NETEXT" ]; then
+    # Whenever the NE bundle is present we re-sign it below with restricted
+    # entitlements. Without an embedded provisioning profile that re-sign
+    # produces a NE that fails activation on SIP-on hosts — the exact bug
+    # the entitlements re-sign exists to fix on the sysext side. Fail-fast
+    # rather than silently shipping the same broken shape.
+    if [ -d "$NETEXT" ]; then
+        if [ ! -f "$PROFILE_NET" ]; then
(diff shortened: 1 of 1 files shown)


getvictor:
@coderabbitai full review

PR #52, opened 2026-04-27
Make reconcile defensive paths reachable to lift Sonar coverage gate

getvictor:
The new-code coverage gate kept slipping below 80% because three "can't actually fail in production" branches in reconcile.go weren't covered by tests:

- emitSyntheticExit's r.newID error (UUID generator failure)
- emitSyntheticExit's json.Marshal error (only fails on non-marshalable types, but we feed it a map[string]any of int+int+string which always marshals cleanly)
- Run's WarnContext on a RunOnce error that RunOnce never returned (signature lied — error was always nil)

Two changes:

1. RunOnce returns just `int` now, not `(int, error)`. Per-PID failures (enqueue / marshal / UUID) are logged inline and never propagate up — there's nothing the caller can do that the inline log path hasn't already done. The Run loop drops its useless error branch.

2. Reconciler gains two test seams: `newID func() (string, error)` defaulting to newUUIDv4, and `marshal func(any) ([]byte, error)` defaulting to json.Marshal. Production callers go through New() which sets both to the production defaults; tests inject failing versions to exercise the error branches.

Two new tests cover the seams:
- TestEmitSyntheticExit_NewIDError
- TestEmitSyntheticExit_MarshalError

Reconcile package coverage 94.8% → 98.6%; only newUUIDv4's rand.Read failure path remains uncovered (one line — crypto/rand on macOS/Linux uses /dev/urandom and doesn't fail).

No behaviour change in production. The wire format, the proctable side-effects, and the per-PID log lines are all identical.

## Release Notes

* **Bug Fixes**
  * Enhanced error handling in reconciliation to gracefully handle individual item processing failures, enabling automatic retries for failed operations instead of terminating the entire pass.

* **Tests**
  * Added test coverage for failure scenarios during UUID generation and payload serialization.

### DIFF
--- a/agent/reconcile/reconcile.go
+++ b/agent/reconcile/reconcile.go
@@ -19,6 +19,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log/slog"
 	"syscall"
 	"time"
@@ -237,7 +238,7 @@ type eventEnvelope struct {
 func (r *Reconciler) emitSyntheticExit(ctx context.Context, hostID string, pid int32, now time.Time) error {
 	id, err := r.newID()
 	if err != nil {
-		return err
+		return fmt.Errorf("generate event id: %w", err)
 	}
 	env := eventEnvelope{
 		EventID:     id,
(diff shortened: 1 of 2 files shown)


getvictor:
@coderabbitai full review

PR #53, opened 2026-04-27
UI polish: 7 demo-blocking findings from claude/mvp/wip.txt

getvictor:
1. TopNav account block was unstyled — "admin@fleet-edr.localLog out" ran together with no visual separation. Add the missing .top-nav__account / __avatar / __user / __logout rules so the user menu sits on the right of the bar (margin-left:auto), with a green avatar circle (initial), monospace email, and an outlined-pill log out button. Pattern matches Fleet's own UI + GitHub / Atlassian conventions.

2. D3 process tree drew links from the synthetic root we add when there are multiple top-level real roots. The synthetic node is filtered out by `pid !== 0` from the rendered nodes set, but its outgoing edges still appeared as ghost lines drifting off-screen left. Filter the links array on the same pid !== 0 check.

3. Search filter previously dimmed non-matches with a CSS opacity class. On a 200-process tree this still leaves 195 dimmed nodes the user has to skim past. Add a queryFilterIds memo that walks the raw tree, keeps every match plus its ancestors, and prunes everything else *before* the d3 layout runs. The dimming classes stay (no-op now since non-matches are gone) so the existing match- highlighting still works without rework.

4. Input placeholders read as default values, not hints. Style ::placeholder explicitly: medium-grey + italic, opacity:1 to defeat Firefox's 0.54 default. Update Policy form placeholders to "e.g. ..." prefix so they're unambiguously hints.

5. Policy save flow had no clear dirty indicator — operators added rows then didn't realise they had to scroll to a separate "Save" section. Replace with a sticky save bar that only appears while dirty, summarising the delta (+1 path, -2 hashes, …), with the audit reason input + Save changes + Discard buttons in line. Matches GitHub's branch-protection editor and AWS Console settings flows. The "policy-flash" status banner replaces the previous one-word "Saved." text after a successful save.

6. ATT&CK coverage was a JSON download only — unsatisfying as a demo prop and harder to read than necessary. New /ui/coverage route renders the same Navigator-layer data as a tactic-grouped table with metric cards (techniques covered / detection rules / tactics with coverage), technique IDs linking to attack.mitre.org, and "Covered by" listing each rule. Keep the JSON download + "Open MITRE Navigator" link in the page header for the procurement workflow. Drop the redundant Download button on the alerts page.

7. Process-tree page had two stacked back arrows ("← Hosts" in the title plus a separate "← Alerts" in the alert breadcrumb) and the alert breadcrumb had a pink/red background that read as an error state. Replace the title's back arrow with a "Hosts / <hostid>" breadcrumb (clickable Hosts) and switch the alert breadcrumb to a neutral grey container — severity is already shown via the badge inside the row, the container itself shouldn't be a status colour.

All 7 verified live via Chrome MCP against the dev server with both hosts (edr-qa + edr-dev) feeding events.

## Summary by CodeRabbit

* **New Features**
  * Added Attack Coverage view displaying detection metrics, technique mappings, and navigator layer download
  * Added Coverage navigation link and user account display in top navigation

* **Improvements**
  * Enhanced Process Tree with query-based filtering and improved breadcrumb navigation
  * Updated Policy Editor with staged edits and explicit save workflow

* **Changes**
  * Removed ATT&CK Navigator download feature from alerts list
  * Various styling refinements throughout the interface

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -44,6 +44,17 @@ sonar.javascript.lcov.reportPaths=ui/coverage/lcov.info
 #                                       first ui/src/**/*.test.{ts,tsx} lands.
 sonar.coverage.exclusions=**/cmd/**/main.go,server/detection/testharness/**,schema/**,extension/edr/**,agent/receiver/**,agent/logging/**,server/bootstrap/**,server/httpserver/**,test/**,ui/src/**
 
+# Exclude static catalog / lookup tables from copy-paste duplication detection.
+# Files where every entry repeats the same property names (id/name/tactic per
+# row, etc.) are 60-90% "duplicated" by Sonar's CPD heuristic, but it's not
+# real code duplication — it's data shape. The new-code duplication gate
+# (3% threshold) goes ERROR on a single such file. Add specific paths here
+# rather than a broad glob so genuinely repetitive code doesn't sneak past.
+#
+#   - ui/src/components/attack-techniques.ts  MITRE ATT&CK technique catalog
+#                                             (id/name/tactic per technique).
+sonar.cpd.exclusions=ui/src/components/attack-techniques.ts
+
 # Skip C/C++/Objective-C analysis (requires a compilation database we don't generate).
 # The only C in the repo is small CGo bridge files in agent/.
 sonar.c.file.suffixes=-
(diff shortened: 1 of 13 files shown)


getvictor:
@coderabbitai full review

PR #54, opened 2026-04-27
UI polish round 2: process tree breadcrumb, coverage column alignment…

getvictor:
…, drop external Navigator link

From claude/mvp/wip.txt rounds 8-10:

1. ProcessTree title was "Hosts / <hostid>", but Hosts is already in the top nav — repeating it inside the page header is noise. Drop the prefix and link the host_id itself to /hosts/<hostid> (no query params), so an operator on an alert-context view (?alert=…) can click out of alert focus to the full host tree without round-tripping through the host list.

2. AttackCoverage page rendered one <Table> per tactic, which let each table size its columns independently from its widest cell — Technique / Name / Covered-by columns drifted by tactic. Switch to a single fixed-layout table with <colgroup> widths (12ch / auto / 35%) and tactic-header rows inside <tbody> via colspan=3. Same pattern Crowdstrike Falcon and Elastic Security use for their ATT&CK coverage tables.

3. "Open MITRE Navigator →" external link was noise: sec ops doesn't need it (the in-product table is now the workflow), and clicking it just opens the Navigator UI with no payload — the operator would have to download the JSON separately and re-upload. Remove. Keep the JSON download but rename to "Export JSON" with a tooltip that calls out the actual audience: procurement / threat-modeling teams who use the upstream Navigator. Sec ops reads everything on the page itself.

## Summary by CodeRabbit

**New Features**
- New "Export JSON" button provides streamlined Attack Coverage data export functionality with explanatory tooltip guidance.

**Style**
- Consolidated Attack Coverage tactic display from multiple sections into a single unified table layout for consistent column alignment.
- Updated Process Tree breadcrumb navigation to provide direct links to individual host detail pages.

### DIFF
--- a/ui/src/components/AttackCoverage.scss
+++ b/ui/src/components/AttackCoverage.scss
@@ -3,23 +3,6 @@
 @use "../styles/var/padding" as *;
 
 .attack-coverage {
-  &__actions {
-    display: flex;
-    align-items: center;
-    gap: $pad-medium;
-  }
-
-  &__navigator-link {
-    color: $core-fleet-green;
-    text-decoration: none;
-    font-weight: $bold;
-    font-size: 0.95em;
-
-    &:hover {
-      text-decoration: underline;
-    }
-  }
-
   &__summary {
     display: flex;
     gap: $pad-large;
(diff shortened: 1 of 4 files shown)


getvictor:
@coderabbitai full review

PR #58, opened 2026-04-28
Add per-rule operator docs (markdown + /ui/rules/<id>)

getvictor:
The detection rules previously had no user-facing documentation outside the source-file headers, which the demo, pilot security review, and the existing TODO in docs/best-practices.md all need. Add structured documentation as part of the rule contract so the markdown reference and the UI rule-detail page render from one source and cannot drift.

- detection.Rule grows a Doc() method that returns a typed Documentation struct (title, summary, description, severity, event_types, false_positives, limitations, config knobs). Each of the eight registered rules implements it. Compile-time gate: a new rule cannot ship without a description.
- detection.RuleMetadata + admin.RuleMetadata carry the Doc through to the admin layer. New endpoint GET /api/v1/admin/rules returns the catalogue as JSON, registered alongside /admin/attack-coverage and gated by the same session+CSRF stack.
- tools/gen-rule-docs walks the catalogue and writes docs/detection-rules.md with an index table, per-rule sections, ATT&CK technique links, configuration table, FP + limitation lists. Output is byte-stable. Wired to `task docs:rules`.
- UI: new /ui/rules/:ruleId route renders the same data via RuleDetail.tsx. AttackCoverage.tsx wraps rule names in the "Covered by" column with <Link> to the detail page. README links the markdown reference from the operator-docs section.

Tests: full integration test that every registered rule has non-empty Doc().Title/Summary/Description/Severity/EventTypes (the gate on new rules). Generator output sanity-checked: every ID appears in the index table and as a section heading; sub-technique URLs use slash, not dot; config env vars surface in the per-rule table. Admin endpoint round-trip + nil-catalog tests cover the new handler. Manually QA'd in Chrome: coverage page links resolve, persistence_launchagent renders all sections, osascript_network_exec renders Critical-severity badge and the no-config branch correctly skips the Configuration section.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive rule documentation system with new admin API endpoint exposing all detection rules, their metadata (severity, ATT&CK mappings, configuration options, false positives, limitations), and new web UI for browsing individual rule details.

* **Documentation**
  * Generated detection rules documentation with rule descriptions, matched event types, and operator guidance.

PR #60, opened 2026-04-28
Doc audit: fix code-vs-doc drift in three spots

getvictor:
Audited every checked-in markdown file (22 files) for accuracy against the current code. Three real drifts found and fixed; everything else verified accurate.

Version drift:
- README.md and docker-compose.prod.README.md showed `EDR_VERSION=v0.5.0` in the production-deployment quickstart, but the project has only ever shipped v0.1.0-rc.* releases. Pinned to `v0.1.0` in three locations (matching the actual release line) so a pilot operator copy-pasting either snippet doesn't try to pull a non-existent image. Also updated the "no DB migration needed within the v0.5.x series" line to v0.1.x to match.

API surface drift:
- docs/api.md's session-cookie endpoint list was missing `GET /api/v1/admin/attack-coverage` (B4) and `GET /api/v1/admin/rules` (the new rule-detail backend added in PR #58). Both routes are registered on the session-protected mux in server/cmd/fleet-edr-server/main.go but were never reflected in the external API doc. Added with one-line descriptions.

Metric-vs-attribute drift:
- docs/operations.md's "core metrics to chart" table listed `edr.policy.fanout_failed` as a counter, but the codebase only emits it as a span attribute and a slog key inside the policy_update audit log line — there is no Int64Counter for it. Moved the guidance out of the metrics table into prose that points operators at the trace UI's failed-span filter, and added the (genuinely-real) `edr.processes.ttl_reconciled` counter that was missing from the table.

Verified accurate (no changes needed):
- All other EDR_* env vars referenced in docs match either server/config/config.go or agent/config/config.go (the agent uses `EDR_ALLOW_INSECURE` while the server uses `EDR_ALLOW_INSECURE_HTTP` by design — both are documented correctly in their respective install guides).
- README badge URLs all point at workflow files that exist.
- Agent install paths (`/etc/fleet-edr.conf`, `/var/db/fleet-edr/...`, `/Library/LaunchDaemons/com.fleetdm.edr.agent.plist`) match agent/config/conffile.go and agent/config/config.go.
- All cross-document relative links resolve.

## Summary by CodeRabbit

* **Documentation**
  * Updated production deployment version reference from v0.5.0 to v0.1.0
  * Added API documentation for ATT&CK Navigator coverage and rule documentation endpoints
  * Added new operational metric for process TTL reconciliation tracking and clarified policy failure reporting


getvictor:
@coderabbitai full review

PR #61, opened 2026-04-28
Move the rule-doc link from alerts list to alert detail breadcrumb

getvictor:
The previous fix put a /ui/rules/<id> link on the alert list's title column, which broke the primary alert pivot — clicking a fired alert no longer opened that alert's process tree, it dumped the analyst onto the rule's documentation page. Wrong workflow.

Restore + relocate:

- ui/src/components/AlertList.tsx
  - Title cell: <Link to="/hosts/<host_id>?alert=...&process=...&at=...">. Single click opens the alert's investigation surface (host process tree pinned to the moment the alert fired with the breadcrumb visible). Same behaviour as before the previous PR.
  - Host cell: <Link to="/hosts/<host_id>"> (no alert pivot params). Distinct destination from the title — title means "this alert", host means "this host". Both are real <a> tags so middle-click and copy-link-address keep working.

- ui/src/components/ProcessTree.tsx
  - Alert breadcrumb's title is now the <Link to="/rules/<rule_id>">. An analyst standing in front of a fired alert can jump straight to "what does this rule do" without losing the process-tree context — the breadcrumb stays put, only the title gains the hover affordance.

- ui/src/components/ProcessTree.scss
  - New &__title--link variant strips the default anchor styling so the breadcrumb still reads as a label, with the green underline appearing only on hover. Visually identical to the previous span-based version when the cursor is elsewhere.

Verified in Chrome:
- /ui/alerts: title → /ui/hosts/<host>?alert=149&process=103463&at=...
- /ui/alerts: host_id → /ui/hosts/<host> (no query string)
- /ui/hosts/<host>?alert=...: breadcrumb title → /ui/rules/credential_keychain_dump

### DIFF
--- a/ui/src/components/AlertList.tsx
+++ b/ui/src/components/AlertList.tsx
@@ -123,29 +123,30 @@ export function AlertList() {
                   </Badge>
                 </td>
                 <td>
-                  {/* Title links to the rule's documentation page so an
-                      analyst seeing an unfamiliar finding can read what the
-                      rule does, why it's almost always malicious, and what
-                      the false-positive sources are without leaving the
-                      app. The host process-tree pivot moved to the Host
-                      column below. */}
+                  {/* Title is the primary alert pivot — it opens the host's
+                      process tree pinned to the moment the alert fired, with
+                      the alert breadcrumb visible. The "what does this rule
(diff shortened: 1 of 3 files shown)


getvictor:
@coderabbitai full review

PR #95, opened 2026-05-02
Phase 5: graph package moved to detection/internal/graph

getvictor:
Adds the Builder + Query under the detection bounded context, both consuming detection/api types and *detection/internal/mysql.Store directly. The forest-building algorithm and the
ingest-time/kernel-time correlation pad are preserved verbatim from server/graph; only the package name and import paths change.

Updated detection/api types to support the move:
- ProcessNode: gained NetworkConnections + DNSQueries to match the existing wire shape that server/graph.ProcessNode emits today (preserving GET /api/hosts/{id}/tree's response bytes).
- ProcessDetail: NEW type matching server/graph.ProcessDetail, the shape GET /api/hosts/{id}/processes/{pid} returns.
- Service.BuildTree: signature corrected to return []ProcessNode (forest) and GetProcessDetail returns *ProcessDetail (full detail with re-exec chain), aligning the api with what the actual graph + operator surface delivers.

The existing server/graph package is unchanged so cmd/main still builds against it; the new package becomes the canonical home and the old one gets removed once cmd/main wiring switches.

Engine + filter scaffolding stays empty; the Engine implementation lands together with the rules.api alias collapse + cmd/main wiring because the three changes are tightly coupled: until rules.api.Event aliases to detection.api.Event, the Engine's call to rule.Evaluate has type mismatches across every rule file.

## Summary by CodeRabbit

* **New Features**
  * Process detail endpoint now includes network connections and DNS query history for enhanced forensic investigation
  * Process tree builder returns multiple root processes per query for greater flexibility
  * Process re-execution chains are captured and available for lifecycle analysis

### DIFF
--- a/server/detection/api/service.go
+++ b/server/detection/api/service.go
@@ -20,8 +20,8 @@ import (
 type Service interface {
 	// Operator reads.
 	ListHosts(ctx context.Context) ([]HostSummary, error)
-	BuildTree(ctx context.Context, hostID string, tr TimeRange, depth, limit int) (ProcessNode, error)
-	GetProcessDetail(ctx context.Context, hostID string, pid int, atTimeNs int64) (*Process, error)
+	BuildTree(ctx context.Context, hostID string, tr TimeRange, limit int) ([]ProcessNode, error)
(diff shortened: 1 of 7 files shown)


getvictor:
@coderabbitai full review

PR #99, opened 2026-05-03
Phase 7: per-context testkit packages; bootstrap is production-only

getvictor:
Introduces server/<X>/testkit/ for each of the five bounded contexts (identity, endpoint, rules, response, detection) as the curated test-fixture surface. Tests reach for testkit; production wiring (cmd/main, server/testdb/full) reaches for bootstrap.

Why: phase 6 left two non-API cross-context allowances in arch-go (**.detection.bootstrap and **.detection.testharness) so the catalog rule tests could apply detection schema and use testharness.Scenario. Both weakened the API-only bounded-context rule. Phase 7 collapses them into a single **.detection.testkit allowance, generalises the pattern across every context, and drops every test's direct dependency on its own context's bootstrap package.

Per-context surface:

- <X>/testkit/testkit.go: ApplySchema (and MigrateSchema for detection). Thin wrappers over bootstrap so the test contract is importable separately from the production wiring contract.
- detection/testkit/scenario.go + replay.go: lifted from detection/testharness/. Same code, new path. Replay also uses testkit's own Apply/Migrate so the runner doesn't reach for detection/bootstrap directly.

testdb/full now composes testkit.Apply across the five contexts (was calling bootstrap.ApplySchema). cmd/main is unchanged — production wiring still lives in bootstrap.

arch-go strict mode tightening:

- Each <X>/internal/* allow-list enumerates **.<X>.{api, internal.**, testkit} — drops the over-broad **.<X>.** that previously let internal code import its own bootstrap. Internal production code cannot call its own context's bootstrap; tests reach for testkit.
- New per-testkit rules pin each testkit to its own context, so cross-context test allowances (rules.internal -> detection.testkit) cannot be exploited as transitive sneak-in paths to a third context. Verified by transient violator files (one direct cross-context import in policy, one transitive sneak-in via testkit) before this commit.
- rules.internal allow-list drops **.detection.bootstrap and **.detection.testharness; gains **.detection.testkit. The non-API cross-context door is closed.

detection/testharness/ is deleted. Catalog rule fixture tests (credential_keychain_dump, privilege_launchd_plist_write, sudoers_tamper, testhelpers) and detection-internal tests all migrate to detection/testkit.

ADR-0004 updated with the testkit/bootstrap split.

go build ./..., task lint:go, task lint:arch, and full server test suite all clean.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized testing infrastructure across bounded contexts with dedicated testkit packages as the public testing surface
  * Enforced stricter module dependency boundaries

* **Tests**
  * Updated test suites to use new testkit infrastructure for database schema initialization

* **Documentation**
  * Updated architecture documentation to reflect module boundary changes and testing structure

### DIFF
--- a/arch-go.yml
+++ b/arch-go.yml
@@ -3,81 +3,160 @@ version: 1
 # Architecture rules for the modular-monolith bounded-context migration.
 # See docs/adr/0004-modular-monolith-bounded-contexts.md for the rationale.
 #
-# Phase 6 strict mode: every transitional allow-list entry from earlier
-# phases (`**.store`, the `**.detection` catch-all) is gone. The five
-# bounded contexts are stable; new violations should fail the build.
+# Phase 7 strict mode: bootstrap is for production wiring only.
+# Per-context internal code may NOT import its own context's bootstrap;
+# tests reach for the per-context testkit instead. Cross-context tests
+# import only **.<Y>.api or **.<Y>.testkit — never **.<Y>.bootstrap and
+# never **.<Y>.internal.**. testkit packages are pinned to their own
+# context so cross-context test allowances cannot be exploited as
+# transitive paths to other bounded contexts.
 
 dependenciesRules:
-  # identity context internals may import only their own context,
-  # platform packages, and standard library / approved third-party.
-  # Other contexts that need identity functionality go through
-  # identity/api.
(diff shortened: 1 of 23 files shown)


getvictor:
@coderabbitai full review

Issue #87, opened 2026-05-02
Append-only operator audit trail (forensic chain of custody)


getvictor:
## Context

Today every operator-initiated state change emits a structured slog
line plus a few OTel span attributes. Examples:

- PUT /api/policy: `edr.admin.action=policy_update`, actor, reason,
  version, fanout counts.
- POST /api/enrollments/{host_id}/revoke: actor, reason, host_id.
- POST /api/commands: admin_action, host_id, command_type, command_id.

That's enough for live operational debugging via SigNoz / Tempo /
Loki. It is NOT enough for forensic chain-of-custody:

1. **Mutability** — slog ships to a log backend that may be edited /
   purged / re-indexed. OTel spans similarly. There's no in-database
   record an investigator can attest to under a discovery request.
2. **Retention coupling** — log backends typically keep 30-90 days.
   IR + insurance-claim windows for security incidents run to
   multi-year. The same engineer who triages an alert today must be
   able to attest to who did what 18 months later.
3. **Tamper-evidence** — a compromised admin can edit logs (or the
   log backend) to cover their tracks. An append-only DB table with
   per-row signatures (HMAC chain or Merkle tree) is harder to
   silently mutate.
4. **Multi-admin attribution** — once the operator pool grows past
   one human, audit needs to be queryable by actor + time range +
   action class. Grep-ing slog files doesn't scale.

CrowdStrike, S1, and Defender all expose forensic audit trails as a
first-class operator surface. SOC2 / ISO 27001 / HIPAA auditors
expect them. PCI-DSS §10.2 explicitly requires it for systems
storing/processing card data.

## Scope when picked up

1. **New table** in (probably) the identity context:

   ```sql
   CREATE TABLE audit_events (
       id BIGINT AUTO_INCREMENT PRIMARY KEY,
       at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
       actor_user_id BIGINT NOT NULL,
       action_class VARCHAR(64) NOT NULL,  -- "policy_update", "command_create", "alert_state_change", "host_revoke", ...
       target_kind  VARCHAR(64) NOT NULL,  -- "policy", "command", "alert", "host"
       target_id    VARCHAR(255) NOT NULL,
       payload      JSON NOT NULL,         -- before/after diff or full body
       prev_chain_hash VARBINARY(32) NULL, -- hash of previous row (Merkle chain)
[...]

Issue #80, opened 2026-05-01
Use scheme-appropriate (or no) WWW-Authenticate header on session-cookie 401s


getvictor:
## Context

Reviewer bots (Gemini Code Assist on PR #78) noted that the operator-session middleware at \`server/identity/internal/middleware/session.go\` sets

\`\`\`
WWW-Authenticate: Bearer error=\"invalid_token\"
\`\`\`

on 401 responses, but the middleware authenticates via the \`edr_session\` HttpOnly cookie, not Bearer tokens. RFC 6750 reserves the \`Bearer\` challenge scheme for Bearer-token auth; for cookie-based auth there is no standardized challenge scheme in RFC 7235.

The same header is correctly emitted by the host-token middleware in \`server/authn/authn.go\` (which IS Bearer auth and stays unchanged after the modular-monolith migration).

## Decision needed

Three options:

1. **Omit \`WWW-Authenticate\` from Session-middleware 401s.** Most accurate to RFC 7235; cookie-based auth has no standard challenge scheme, so omitting is the correct \"unknown\" behavior. Clients ignore the header on cookie auth anyway because they have no credential to retry with.
2. **Use a custom scheme** (e.g., \`Cookie\`). Non-standard, may confuse clients.
3. **Keep \`Bearer\`** (current). Wrong per RFC, but harmless because no client treats this as actionable for cookie auth.

## Why this isn't fixed in PR #78

PR #78 (modular-monolith phase 1) explicitly commits to wire-compatibility with main. Changing the header on 401 responses, even toward more correct behavior, would be a wire-shape change inappropriate for that PR's scope.

## Suggested approach

A small follow-up PR that:
- Removes the \`WWW-Authenticate\` header from \`identity/internal/middleware/session.go\`'s \`fail\` function (or splits the helper so cookie-flavored 401s use a different code path).
- Updates \`server/authn/authn.go\` to keep emitting it for the host-token path (Bearer auth there is correct).
- Adds tests asserting the header is present on host-token 401s and absent on session 401s.

## References
- RFC 7235 §4.1 (WWW-Authenticate)
- RFC 6750 §3 (Bearer challenge)
- PR #78 review comment: https://github.com/getvictor/fleet-edr/pull/78

getvictor:
Landed in #108 (merged 2026-05-03 as 3c70a33).

server/httpserver/authfail.go now exposes two writers:
[...]

PR #108, opened 2026-05-03
Add operator audit trail (#87); drop Bearer challenge on cookie-auth …

getvictor:
…401s (#80)

Both fixes get more expensive every day they don't ship. The audit trail opens a permanent gap in the historical record the moment the server serves operator actions without it; the misleading Bearer challenge on cookie-protected endpoints confuses browser auth flows and HTTP clients that key retry behaviour off WWW-Authenticate.

Audit (#87):
  - audit_events table (occurred_at, actor_user_id, actor_email, action, target_type, target_id, trace_id, remote_addr, payload JSON), four indexes covering the retrieval queries the admin UI will run.
  - identity/api.AuditRecorder and AuditReader interfaces, plus 9 AuditAction constants namespaced <resource>.<verb>. Append-only is enforced by the interface shape: no Update or Delete methods, no UPDATE/DELETE paths in the audit package.
  - MySQL Store + GET /api/audit retrieval handler under identity/internal/audit. trace_id is pulled from ctx so audit rows pivot to SigNoz spans by the same handle that lands on the response X-Request-ID header.
  - Wired at five call sites: login/logout/login_failed in identity, alert.{acknowledge,resolve,reopen} in detection, policy.update in rules, command.issue in response, enrollment.revoke in endpoint. Each call site soft-fails on Recorder errors (the action committed; a missing audit row is a follow-up incident, not a reason to fail an HTTP response that already returned 200/204/201).
  - arch-go widened to allow rules.internal and endpoint.internal to import identity.api for the AuditRecorder.

WWW-Authenticate (#80):
  - Split httpserver.WriteAuthFailure into Bearer (existing semantics, still emits the RFC 6750 challenge on 401 for agent endpoints) and new WriteCookieAuthFailure (no header, ever). The cookie variant documents in its doc comment why RFC 7235 has no scheme for cookie auth and inventing one is worse than omitting the header.
  - identity/middleware/session.go's seven callers (Session + CSRF) now use the cookie variant; hosttoken middleware in endpoint stays on the Bearer variant.

Folded-in dev-environment fixes:
  - Taskfile.yml dev:server: empty-password DSN (matches docker-compose MYSQL_ALLOW_EMPTY_PASSWORD; root:toor was failing access-denied), bind 0.0.0.0:8088 so a UTM VM agent on the bridge interface can reach it, OTel envs declared (parent shell still wins on conflict per Taskfile v3 semantics).
  - New ui:setup task with a status guard runs npm ci on a fresh worktree, wired as deps of build:ui and dev:ui so a clean clone builds without separate setup.

Tests added:
  - authfail_test.go covers both Bearer (challenge on 401, omitted on 5xx) and Cookie (no challenge on any status) writers, plus the shared body/log shape.
  - audit/mysql_test.go covers Record round-trip, login_failed without a users row (denormalised actor_email read path), trace_id capture from ctx, action filter, BeforeID pagination, empty-action rejection, nil-db panic, and stable AuditAction constants.
  - session_test.go regression assertion: cookie-auth 401 carries no WWW-Authenticate.

Verified end-to-end against the dev server: nine distinct AuditAction values land in audit_events with correct actor, target, payload, and trace_id matching the dev-server access-log trace_id for the same request. Chrome login + UI policy save produced an audit row with the operator's email and the supplied reason in the payload.

## Release Notes

* **New Features**
  * Added comprehensive audit logging for operator actions (login, logout, alert management, policy operations, command issuance, and enrollment operations)
  * Added audit trail search and filtering endpoint with pagination support

* **Improvements**
  * Enhanced development server observability with OTLP exporter configuration
  * Automated UI dependency setup in development workflow
  * Improved authentication failure response handling for cookie-based sessions

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -52,12 +52,18 @@ tasks:
 
   ui:setup:
     desc: |
-      Install UI npm dependencies. Idempotent: skips when ui/node_modules already exists,
-      so it is cheap to wire as a dep of build:ui and dev:ui. A fresh clone or a new git
-      worktree therefore builds without a separate manual `cd ui && npm ci` step.
+      Install UI npm dependencies. Wired as a dep of build:ui and dev:ui so a fresh clone
+      or new git worktree builds without a separate manual `cd ui && npm ci` step. Skips
+      when `node_modules/.package-lock.json` (the snapshot npm writes on a successful
+      `npm ci`) is at least as new as `package-lock.json`, so a lockfile bump triggers a
+      fresh install but a no-op build does not pay the cost. A bare `test -d node_modules`
+      guard would leave stale dependencies in place after a branch switch that touches
+      the lockfile.
     dir: ui
-    status:
-      - test -d node_modules
+    sources:
+      - package-lock.json
+    generates:
+      - node_modules/.package-lock.json
     cmds:
       - npm ci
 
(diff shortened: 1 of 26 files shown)


getvictor:
@coderabbitai full review

Issue #94, opened 2026-05-02
Recursive CTE for GetExecChain (replace bounded N+1 walk)


getvictor:
## Summary

`detection/internal/mysql/processes.go` `GetExecChain` walks the
`previous_exec_id` linkage backward one row at a time, paying a
round-trip per chain link. The loop is bounded at maxChainLen=64,
so the worst case is 64 sequential queries. Gemini Code Assist
flagged this on PR #90 as a **MEDIUM** performance issue.

## Current shape

```go
const maxChainLen = 64
var chain []api.Process
prevID := current.PreviousExecID
for depth := 0; prevID != nil && depth < maxChainLen; depth++ {
    var p api.Process
    err := s.db.GetContext(ctx, &p, `
        SELECT id, host_id, pid, ppid, ...
        FROM processes WHERE id = ? AND host_id = ?`, *prevID, current.HostID)
    if errors.Is(err, sql.ErrNoRows) { break }
    if err != nil { return nil, ... }
    chain = append(chain, p)
    prevID = p.PreviousExecID
}
slices.Reverse(chain)
return chain, nil
```

## Proposed fix (MySQL 8.0+ recursive CTE)

```sql
WITH RECURSIVE chain AS (
  SELECT * FROM processes WHERE id = ? AND host_id = ?
  UNION ALL
  SELECT p.* FROM processes p
  JOIN chain c ON p.id = c.previous_exec_id AND p.host_id = c.host_id
)
SELECT id, host_id, pid, ppid, path, args, uid, gid, code_signing, sha256,
       fork_time_ns, fork_ingested_at_ns, exec_time_ns, exit_time_ns,
       exit_ingested_at_ns, exit_reason, exit_code, previous_exec_id
FROM chain
ORDER BY fork_time_ns
LIMIT 64
```

Single round-trip; rows arrive ordered chronologically (no
slices.Reverse needed). LIMIT 64 preserves the cycle-protection cap
(belt-and-suspenders: schema invariant says
previous_exec_id is strictly decreasing, so a cycle shouldn't be
possible — but a corrupt FK shouldn't lock up the query either).

Verify MySQL version baseline; we already require 8.0+ per
docs/adr/0002 / phase 5 plan.

## Out of scope for phase 5

Pre-existing pattern relocated as-is per the plan's "pure relocation"
intent.

## Acceptance

- One ExecContext / GetContext per `GetExecChain` call.
- Existing test coverage (today: `store/process_reconcile_test.go`,
[...]

Issue #93, opened 2026-05-02
Bulk alert_events linking inside InsertAlert (replace per-event INSERT loop)


getvictor:
## Summary

`detection/internal/mysql/alerts.go` `InsertAlert` runs one INSERT
per event_id when linking the alert to its triggering events. The
loop is inside a transaction (so atomicity is fine) but every event
ID still costs one round-trip. Alerts with many correlated events
(common for noisy detections) pay N latency hits per insert. Gemini
Code Assist flagged this on PR #90 as a **MEDIUM** performance
issue.

## Current shape

```go
for _, eid := range eventIDs {
    if _, err := tx.ExecContext(ctx,
        \"INSERT INTO alert_events (alert_id, event_id) VALUES (?, ?)\",
        alertID, eid); err != nil {
        return 0, false, fmt.Errorf(...)
    }
}
```

Same pattern in `attachEventsToExistingAlert` (the dedup branch),
which uses `INSERT IGNORE` per event.

## Proposed fix

Build a single multi-row INSERT:

```go
const stmt = \"INSERT INTO alert_events (alert_id, event_id) VALUES \"
ph := make([]string, len(eventIDs))
args := make([]any, 0, len(eventIDs)*2)
for i, eid := range eventIDs {
    ph[i] = \"(?, ?)\"
    args = append(args, alertID, eid)
}
_, err := tx.ExecContext(ctx, stmt+strings.Join(ph, \",\"), args...)
```

Same shape for the dedup path with `INSERT IGNORE`.

Cap the batch size at \\~500 placeholders if any caller starts
linking thousands of events (today's rules cap at \\~10).

## Out of scope for phase 5

Pre-existing pattern relocated as-is per the plan's
\"steady-state per-batch cost: identical to today\" constraint.

## Acceptance

- One ExecContext per dedup-branch and one per happy-path branch
  regardless of len(eventIDs).
- Existing tests pass.
- The dedup-skip path still uses INSERT IGNORE semantics so a
  retried link doesn't blow up on `(alert_id, event_id)` PK.

Surfaced by: Gemini Code Assist on PR #90.

getvictor:
Fixed by #110 (merged 2026-05-03). InsertAlert + attachEventsToExistingAlert now use a single chunked multi-row INSERT (with INSERT IGNORE for the dedup branch), capped at 500 rows per statement. Bench: 10 vs 1000 events linked grew ~5.5x (was 100x).

Issue #92, opened 2026-05-02
Indexed PID lookup on events (replace JSON_EXTRACT scan in GetNetworkEventsForProcess)


getvictor:
## Summary

`detection/internal/mysql/processes.go` `GetNetworkEventsForProcess`
filters network events on `JSON_EXTRACT(payload, '\$.pid') = ?`.
That predicate forces a full scan of the `events` table because the
JSON path expression isn't indexable. The events table is the single
largest, hottest table in the system; the scan cost grows linearly
with retained events. Gemini Code Assist flagged this on PR #90 as a
**HIGH** performance issue.

## Current shape

```sql
SELECT event_id, host_id, timestamp_ns, ingested_at_ns, event_type, payload
FROM events
WHERE host_id = ? AND event_type IN ('network_connect', 'dns_query')
  AND ingested_at_ns >= ? AND ingested_at_ns <= ?
  AND JSON_EXTRACT(payload, '\$.pid') = ?
ORDER BY timestamp_ns
```

## Proposed fix (two options)

**Option A — generated/virtual column + index** (preferred, MySQL 5.7+):

```sql
ALTER TABLE events
  ADD COLUMN payload_pid INT GENERATED ALWAYS AS
    (CAST(JSON_UNQUOTE(JSON_EXTRACT(payload, '\$.pid')) AS UNSIGNED)) STORED,
  ADD INDEX idx_events_host_type_pid (host_id, event_type, payload_pid, ingested_at_ns);
```

Query becomes:
```sql
WHERE host_id = ? AND event_type IN ('network_connect', 'dns_query')
  AND payload_pid = ?
  AND ingested_at_ns >= ? AND ingested_at_ns <= ?
```

STORED so the index is range-scannable; VIRTUAL would force per-row
re-evaluation. Backfill is automatic on ALTER for STORED columns.

**Option B — functional index** (MySQL 8.0.13+):

```sql
ALTER TABLE events ADD INDEX idx_events_pid_path
  ((CAST(JSON_UNQUOTE(JSON_EXTRACT(payload, '\$.pid')) AS UNSIGNED)));
```

Less invasive (no new column) but requires the query to express the
exact same JSON expression to hit the index.

## Out of scope for phase 5

Phase 5 is structurally a relocation; schema changes beyond the FK
drop are explicitly out of scope per phase5.md. This issue captures
the optimization for a dedicated schema-tier follow-up.

## Acceptance

[...]

Issue #91, opened 2026-05-02
Bulk hosts upsert in UpsertHosts (replace per-host INSERT loop)


getvictor:
## Summary

`detection/internal/mysql/hosts.go` `UpsertHosts` executes one
`INSERT ... ON DUPLICATE KEY UPDATE` per distinct host in the batch.
For an event batch spanning N hosts that's N round-trips inside the
ingest hot path. Gemini Code Assist flagged this on PR #90 as a
**HIGH** performance issue.

## Current shape

```go
for hostID, st := range byHost {
    _, err := s.db.ExecContext(ctx, `
        INSERT INTO hosts (host_id, event_count, last_seen_ns)
        VALUES (?, ?, ?)
        ON DUPLICATE KEY UPDATE
            event_count = event_count + VALUES(event_count),
            last_seen_ns = GREATEST(last_seen_ns, VALUES(last_seen_ns))`,
        hostID, st.count, st.maxTSNs)
    ...
}
```

## Proposed fix

Single batched statement with multi-row VALUES clause:

```go
const tmpl = `INSERT INTO hosts (host_id, event_count, last_seen_ns)
VALUES %s
ON DUPLICATE KEY UPDATE
  event_count = event_count + VALUES(event_count),
  last_seen_ns = GREATEST(last_seen_ns, VALUES(last_seen_ns))`
```

Build the placeholders + args once, fire one ExecContext.

## Out of scope for phase 5

Phase 5's plan explicitly preserves the existing per-batch SQL shape
(*"Steady-state per-batch cost: identical to today"*). This issue
captures the win for a follow-up perf pass once the modular monolith
migration completes.

## Acceptance

- One `ExecContext` per `UpsertHosts` call regardless of N.
- Existing test coverage (`store/host_test.go`, then
  `detection/internal/mysql/hosts_test.go` after the test relocation)
  passes unchanged.
- Microbenchmark: per-batch upsert latency drops linearly with N.

Surfaced by: Gemini Code Assist on PR #90.

getvictor:
Fixed by #110 (merged 2026-05-03). UpsertHosts now uses a single multi-row INSERT regardless of host count; benchmark shows 16 vs 256 hosts within ~2x (was linear per the prior shape).

PR #110, opened 2026-05-03
perf(detection): bulk SQL + indexed PID + recursive exec chain

getvictor:
Four related performance improvements in server/detection/internal/mysql/, all surfaced by Gemini Code Assist on PR #90:

#91 (HIGH) hosts.go UpsertHosts: per-host INSERT loop replaced by a single multi-row INSERT ... ON DUPLICATE KEY UPDATE. N round-trips inside the ingest hot path collapse to one regardless of distinct-host count.

#92 (HIGH) processes.go GetNetworkEventsForProcess: JSON_EXTRACT predicate (forces full scan of events) replaced by a STORED generated column payload_pid backed by idx_events_host_type_pid_ingested. The new test asserts EXPLAIN reports the new composite index. Schema added in detection/bootstrap (CREATE TABLE for fresh installs + idempotent ALTER for existing DBs via the IgnoreErrors pattern).

#93 (MED) alerts.go InsertAlert + attachEventsToExistingAlert: per-event INSERT loops replaced by a single multi-row INSERT (and INSERT IGNORE for the dedup path). Chunked at 500 rows/statement so a future noisy detection that links thousands of events doesn't breach max_allowed_packet.

#94 (MED) processes.go GetExecChain: 64-step backward walk replaced by a WITH RECURSIVE CTE. Single round-trip; depth column inside the CTE provides belt-and-suspenders cycle protection alongside the schema's strictly-decreasing previous_exec_id invariant. Returns chain oldest-first directly (slices.Reverse no longer needed).

Microbenchmarks in perf_test.go validate the linear-with-N drop the issues' acceptance criteria call for: BenchmarkInsertAlert grows ~5.5x between 10 and 1000 events linked (was 100x); BenchmarkGetExecChain is near-flat across depths 8/32/64.

testdb.Open widened from *testing.T to testing.TB so benchmarks share the same per-test-isolated DB fixture; every method used (Helper, Fatalf, Name, Cleanup, Context) is on TB.

### DIFF
--- a/server/detection/bootstrap/migrations.go
+++ b/server/detection/bootstrap/migrations.go
@@ -106,6 +106,24 @@ var migrations = []migrationStep{
 		SQL:          `ALTER TABLE processes ADD INDEX idx_processes_previous_exec (previous_exec_id)`,
 		IgnoreErrors: []uint16{mysqlDuplicateKey, mysqlDuplicateKeyAlt},
 	},
+	// Stored generated column + composite index so GetNetworkEventsForProcess
+	// can drop its JSON_EXTRACT predicate (issue #92). STORED (not VIRTUAL)
+	// because VIRTUAL would force a per-row JSON_EXTRACT re-evaluation on
+	// every index lookup, defeating the whole point. CAST AS UNSIGNED matches
+	// pid's non-negative domain; the column is BIGINT so a JSON pid value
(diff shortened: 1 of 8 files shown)


getvictor:
@coderabbitai full review

Issue #86, opened 2026-05-02
Host-token rotation: rotate agent bearer tokens on a schedule, not enroll-once-forever


getvictor:
## Context

When an agent enrolls today, the server mints a bearer token and the
agent stores it indefinitely (`/var/db/fleet-edr/enrolled.plist`). The
token is rotated only on:

- explicit operator revoke (`POST /api/enrollments/{host_id}/revoke`),
  forcing a re-enrol;
- agent re-enrol on `401_from_server` (which today only happens when an
  operator revokes or a fresh DB doesn't recognise the token).

So in steady-state production, a single token is good for the lifetime
of the host installation. For a top-tier EDR that's a meaningful
security gap:

- An exfiltrated token grants persistent host impersonation. The
  attacker can ingest spoofed events (poisoning the alert pipeline) or
  simply silence the host (acking commands without applying them) until
  somebody notices and manually revokes.
- No periodic re-attestation. Compromise is silent.
- Rotation is the standard mitigation across CrowdStrike, S1,
  Defender — they re-issue per session / per hour / per day depending
  on configuration.

## Scope when picked up

1. **Server side** — extend `server/endpoint/api.Service`:
   - `RotateToken(ctx, hostID) (newToken string, err error)` mints a
     new token, persists the new hash, marks the previous token's
     hash for grace-period acceptance.
   - Schedule policy: every N hours OR every M agent-poll cycles
     (configurable via `EDR_HOST_TOKEN_LIFETIME`, default 24h).
   - Grace window: accept the previous token for ~5 minutes after
     rotation so an in-flight agent request doesn't 401 mid-poll.
   - Audit log on every rotation (`edr.host.token_rotation`).

2. **Agent side** — extend the wire shape:
   - The `set_blocklist` command pattern is host-token-poll-driven;
     we should add a `rotate_token` command type that the server
     queues when it decides to rotate. Agent applies → writes the new
     token to disk → next poll uses it.
   - Alternative: piggy-back on every successful host-token request,
     return a `X-EDR-NEW-TOKEN` header sometimes. Less explicit but
     no command-queue traffic; harder to audit. Probably reject.

3. **Schema** — add `enrollments.previous_token_hash` (NULLABLE
   VARBINARY) + `previous_token_expires_at` (NULLABLE TIMESTAMP) so
   the verifier can match either token during the grace window.

4. **Operator surface** — `POST /api/enrollments/{host_id}/rotate`
[...]

PR #111, opened 2026-05-03
#86 phase 1: schema + storage primitives for host-token rotation

getvictor:
Lays the storage groundwork for host-token rotation without changing the public api.Service surface yet, so this commit can land + soak through CI before the service-layer / operator-endpoint / agent changes pile on. The endpoint context's Verify still returns the same (string, error) shape; new metadata-returning + rotation-committing methods sit alongside it for phase-2 callers.

Schema (additive, idempotent):
- enrollments.host_token_issued_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6). Tracks per-token issuance separately from the per-enrollment enrolled_at, so rotation eligibility is a function of the active token's age, not the host's age.
- enrollments.previous_host_token_id / _hash / _salt (all NULLABLE). When rotation flips, the prior (id, hash, salt) is captured here so the verify path can fall through to the previous token during the grace window.
- enrollments.previous_token_expires_at TIMESTAMP(6) NULL. Bounds the grace window per #86's contract (~5 min default, set by the rotate call site).
- INDEX idx_enrollments_prev_token (previous_host_token_id) so the fall-through verify is also O(1).
- A schemaMigrations array (mirroring identity/bootstrap's pattern) + isAlreadyAppliedMigration helper for the upgrade path: an existing enrollments table gets the same columns ALTERed in, with a backfill UPDATE that copies enrolled_at into host_token_issued_at so any pre-rotation host gets flagged "rotation due" on its next verify rather than starting fresh from the moment of deploy.

Storage (server/endpoint/internal/mysql/store.go):
- VerifyResult struct: HostID, CurrentTokenID (the SHA-256 the caller used; phase-2's RotateToken needs it as the optimistic-lock key), TokenIssuedAt (rotation-eligibility input), and MatchedPrevious (true when the verify hit the previous-token grace path so phase-2 knows another rotation is in flight and must not trigger a fresh one).
- VerifyWithMeta returns the full struct; Verify is now a thin wrapper preserving the existing (hostID, error) return so phase-1 doesn't ripple through identity / detection callers.
- Verify path is split into verifyAgainstCurrent + verifyAgainstPrevious helpers so the grace fallback is a separate, individually-testable branch. Two SQL round-trips on a miss (current then previous), one on a hit; argon2id only runs against the row that was actually found.
- RotateHostToken does the atomic flip in one SQL statement. SET clause ordered previous_* before host_* because MySQL evaluates assignments left-to-right and uses the new value for subsequent reads; reversed order would copy the new id into previous_host_token_id. The UPDATE is gated on host_token_id = currentTokenID so two concurrent rotations serialise: only the one whose currentTokenID still matches commits, the rest get RowsAffected=0 -> ErrRotateRaced.
- ErrRotateRaced sentinel for the lost-race branch; ErrTokenMismatch unchanged.

Tests (server/endpoint/internal/mysql/rotation_test.go):
- TestRotateHostToken_HappyPath: rotation returns a fresh raw token + non-empty previous-id prefix; new token verifies as current; old token still verifies during grace and surfaces MatchedPrevious=true (the contract phase-2's verify-time trigger relies on).
- TestRotateHostToken_OldTokenRejectedAfterGrace: 50ms grace + 300ms sleep proves previous_token_expires_at is the binding cutoff; without this a leaked token would stay valid forever.
- TestRotateHostToken_RaceLoses: a stale currentTokenID surfaces ErrRotateRaced via the optimistic lock.
- TestRotateHostToken_ConcurrentVerifiesProduceOneRotation: 16 goroutines, exactly one commits, the rest race-lose. This is the property the verify-time auto-rotate will rely on under burst poll traffic.
- TestRotateHostToken_RejectsRevoked: rotation against a revoked enrollment must surface as ErrRotateRaced rather than silently rehydrating the row.
- TestRotateHostToken_RejectsBadInputs: empty hostID / empty currentTokenID / non-positive grace fail at the boundary, not inside the SQL.
- TestRotateHostToken_OverwritesPriorPrevious: double rotation while a prior grace is still active discards the original token; the oldest-token-rejected outcome is documented in the test as the acceptable failure mode for a host that missed two rotation windows.

## Summary by CodeRabbit

* **New Features**
  * Added automatic and operator-initiated host token rotation with configurable grace periods for token transition
  * Tokens can now be renewed with a grace window allowing verification of both current and previous tokens

* **Chores**
  * Added configuration parameters `EDR_HOST_TOKEN_LIFETIME` (default 24h) and `EDR_HOST_TOKEN_GRACE` (default 5m) for token rotation management

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -283,12 +283,19 @@ tasks:
       service (MYSQL_ALLOW_EMPTY_PASSWORD=yes); a `root:toor@...` DSN will fail with
       access-denied. OTel exporter targets the local signoz-otel-collector on 4317; if
       no collector is running, exporter init logs warnings and request handling continues.
+
+      Host-token rotation defaults (issue #86): EDR_HOST_TOKEN_LIFETIME 24h,
+      EDR_HOST_TOKEN_GRACE 5m. Override either to a short value (e.g.
+      EDR_HOST_TOKEN_LIFETIME=30s) when QA-ing the rotation flow against a real VM agent
+      so you can observe a rotation cycle inside one minute rather than next-day.
     env:
       EDR_DSN: "root:@tcp(127.0.0.1:3316)/edr?parseTime=true"
       EDR_ENROLL_SECRET: dev-enroll-secret
       EDR_ALLOW_INSECURE_HTTP: "1"
       EDR_LISTEN_ADDR: "0.0.0.0:8088"
       EDR_LOG_FORMAT: text
+      EDR_HOST_TOKEN_LIFETIME: 24h
+      EDR_HOST_TOKEN_GRACE: 5m
       OTEL_EXPORTER_OTLP_ENDPOINT: http://127.0.0.1:4317
       OTEL_EXPORTER_OTLP_PROTOCOL: grpc
       OTEL_EXPORTER_OTLP_INSECURE: "true"
--- a/agent/cmd/fleet-edr-agent/main.go
+++ b/agent/cmd/fleet-edr-agent/main.go
@@ -262,12 +262,13 @@ func startCommander(
 		return
 	}
 	cmdr := commander.New(commander.Config{
-		ServerURL:    serverURL,
-		TokenFn:      tokenProvider.Token,
-		OnAuthFail:   tokenProvider.OnUnauthorized,
-		HostID:       hostID,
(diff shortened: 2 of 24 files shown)


getvictor:
@coderabbitai full review

Issue #81, opened 2026-05-01
Honour X-Forwarded-For from trusted proxies in rate-limit + audit IP detection


getvictor:
## Context

Multiple reviewer bots (Gemini Code Assist high-severity, repeated across PR
#78 and likely on every future PR that touches the auth surface) flag that
rate limiting and audit logging derive the client IP solely from
\`r.RemoteAddr\` via \`server/identity/internal/login/handler.go\`'s
\`remoteIP\` helper. When the server runs behind a reverse proxy or load
balancer, every request appears to come from the proxy IP, so the per-IP
rate limiter collapses all traffic into a single bucket and one user
hitting the limit can lock out everyone behind the same proxy.

Best-practices checklist item §3 (\`docs/best-practices.md\`) acknowledges
this:

> [x] Refusal of \`X-Forwarded-For\` until a trusted-proxy allowlist exists

That mark is the deliberate-decision-for-now: blindly trusting
\`X-Forwarded-For\` is worse than ignoring it, because any client can
spoof it. The proper fix is to respect it only from configured trusted
proxies.

## Why this isn't fixed yet

Production deployment topology hasn't been settled. Different paths have
different proxy stories:

- **Fleet-managed install** (one-line installer puts the server behind
  whatever Fleet's deployment chooses)
- **Direct k8s deployment** (operator picks ingress, may already terminate
  TLS at a Layer 7 proxy)
- **Bare-metal docker-compose** (no proxy)

Each implies different defaults and configuration burden.

## Scope when picked up

1. Add config option \`EDR_TRUSTED_PROXIES\` (CSV of CIDR ranges, e.g.
   \`10.0.0.0/8,172.16.0.0/12\`).
2. Update \`remoteIP\` to:
   - Default: ignore \`X-Forwarded-For\`, return \`r.RemoteAddr\` host portion (current behavior).
   - When \`r.RemoteAddr\` matches a trusted-proxy CIDR: walk the
     \`X-Forwarded-For\` header right-to-left, skipping every entry that
     is itself in the trusted-proxy set, and return the first non-trusted
     IP found. Standard pattern; covers chained proxies safely.
3. Same logic in any other place we use the client IP (audit log,
   enrollment rate limit if it has one, eventually agent ingest).
4. Tests covering: no header (use RemoteAddr), header from untrusted
   source (ignore), header from trusted source (parse), spoofed header
   chain (walk past trusted entries).
[...]

PR #113, opened 2026-05-03
feat(httpserver): trusted-proxy X-Forwarded-For (#81)

getvictor:
Adds a ClientIPResolver that honours X-Forwarded-For only when the direct TCP peer (r.RemoteAddr) is itself a configured trusted proxy. The default — empty trusted list — matches the prior secure stance: XFF is ignored and every IP-keyed surface (rate limiter, audit log) sees the direct peer. Configure via EDR_TRUSTED_PROXIES (CSV of CIDRs or bare IPs) the moment fleet-edr-server runs behind an ALB / nginx / Cloudflare; without it, one user hitting the rate limit locks out everyone behind the proxy.

Resolution semantics (issue #81 spec + OWASP guidance):
  - r.RemoteAddr untrusted -> ignore XFF, return peer IP.
  - r.RemoteAddr trusted   -> walk XFF right-to-left, skip trusted
                              hops, return the first non-trusted entry;
                              fall back to peer if every entry is
                              trusted or the chain is empty.
The right-to-left walk handles chained proxies (ALB -> nginx ->
fleet-edr-server) safely. The leftmost entry — what the original
client claimed — is never treated as inherently trustworthy.

Wire-up:
  - server/httpserver/clientip.go: ClientIPResolver, Middleware that stashes the resolved IP on ctx, and ClientIP(r) helper that reads it (falls back to RemoteIP when the middleware didn't run, so a forgotten wire-up degrades to today's behaviour rather than silently honouring spoofed XFF).
  - server/httpserver/httpserver.go: Options.ClientIPResolver opt-in, middleware installed inside the otelhttp span and outside the access log.
  - server/config: EDR_TRUSTED_PROXIES env var (CSV; bare IPs auto- promoted to /32 or /128 by the resolver).
  - server/cmd/fleet-edr-server/main.go: build resolver from config, pass into httpserver.Build, log the trusted set on startup.

Consumer call-site swaps (RemoteAddr / RemoteIP / local remoteIP -> httpserver.ClientIP):
  - identity/internal/login/handler.go: rate-limit bucket key + audit.
  - endpoint/internal/enroll/handler.go: rate-limit bucket key + audit. Drops the local remoteIP helper and trimHost field — the central helper handles the host-port-stripping cases the old flag toggled.
  - detection/internal/operator/handler.go: alert audit RemoteAddr.
  - endpoint/internal/operator/handler.go: enrollment-revoke audit.
  - response/internal/operator/handler.go: command-issue audit.
  - rules/internal/operator/handler.go: policy-update audit.

Tests:
  - clientip_test.go covers the matrix in the issue spec: no XFF, untrusted-peer ignore, trusted-peer single + multi-hop chain, all-trusted fallback, multiple XFF header instances, malformed + empty entries, IPv6, ::ffff:IPv4-mapped IPv6, bare-IP CIDR. Plus nil-resolver and middleware-ctx wiring.
  - Existing rate-limit tests in login + enroll continue to pass; the httpserver.ClientIP fallback path keeps tests working without rewiring middleware.

## New Features
* Added trusted proxy configuration support to accurately detect client IP addresses when the server operates behind reverse proxies or load balancers. Supports comma-separated IP addresses and CIDR network blocks.

## Refactor
* Updated client IP detection across request handlers to use a unified, proxy-aware resolution mechanism, ensuring consistent client identification in audit logs and rate limiting throughout the system.

### DIFF
--- a/server/cmd/fleet-edr-server/main.go
+++ b/server/cmd/fleet-edr-server/main.go
@@ -150,7 +150,16 @@ func run() error {
 	go runDetection(ctx, detectionCtx, logger)
 	go runIdentity(ctx, identityCtx, logger)
 
-	srv := newHTTPServer(cfg, mux, logger)
+	clientIPResolver, err := httpserver.NewClientIPResolver(cfg.TrustedProxies)
+	if err != nil {
+		logger.ErrorContext(ctx, "EDR_TRUSTED_PROXIES invalid", "err", err)
+		return err
+	}
+	if len(cfg.TrustedProxies) > 0 {
+		logger.InfoContext(ctx, "trusting X-Forwarded-For from proxies", "trusted", cfg.TrustedProxies)
+	}
+
+	srv := newHTTPServer(cfg, mux, logger, clientIPResolver)
 	if err := configureTLSIfEnabled(ctx, logger, srv, cfg); err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


getvictor:
@coderabbitai full review

PR #126, opened 2026-05-06
Feat/user management phase 5a

getvictor:
## Release Notes

* **New Features**
  * Configurable session timeouts with separate idle and absolute limits for standard and break-glass sessions
  * Reauthentication flow enabling users to refresh credentials when accessing sensitive operations
  * Session freshness tracking to determine when re-authentication is required
  * Critical severity alert resolution now requires fresh session authentication


getvictor:
@coderabbitai full review

PR #129, opened 2026-05-08
phase 6: lock chokepoint coverage + ship-v1 operator runbook

getvictor:
The spec frames Phase 6 as a single-PR flip of authz.shadow_mode from true to false. In this codebase that flip already happened in Phase 2 wiring — server/config/config.go reads
EDR_AUTHZ_SHADOW_MODE as opt-in, default false. Phase 6 here is ship-v1 readiness instead: lock coverage so a v1 tag can't carry a privileged handler that silently bypasses the chokepoint, and write the doc the first-deploying operator reads when a 403 lands.

Tests:
- TestAllow_EveryRegisteredActionGrantedSomewhere (server/identity/internal/authz/engine_test.go) iterates RegisteredActions() and asserts each is granted by at least one seeded role. Catches an action added to the registry without a matching grant in roles.json — that action would otherwise ship as a permanent no_matching_rule deny for every caller.
- TestEveryPrivilegedHandlerCallsHTTPGate (test/arch/chokepoint_coverage_test.go) parses every operator handler file under {detection,rules,response,endpoint}/internal/operator/
  + identity/internal/audit/, finds functions with the (http.ResponseWriter, *http.Request) signature, and asserts the file references HTTPGate. Path-based gateExceptions allowlist for files that legitimately don't gate (empty today). Defensive handlerFilesScanned floor catches a parser regression that would otherwise turn the test into a silent no-op.

Boot log:
- main.go now logs `authz enforcement posture=ENABLED|SHADOW` on startup. A typo on EDR_AUTHZ_SHADOW_MODE surfaces immediately rather than via a confused 403 hours later. Separate event from the broader "fleet-edr-server starting" line so a SigNoz / log- shipper alert pivoting on the posture field doesn't have to parse it out of noisier multi-field events.

Comment cleanup:
- config.go's AuthzShadowMode comment dropped the "wave-1 rollout knob" framing — wave 1 IS what v1 ships. Documents the field as the audit-only knob and the boot-only read semantics.

Operator runbook:
- New docs/authz.md. Covers the seeded role matrix (mirroring policy/data/roles.json), the wire shape of a chokepoint deny including the reauth_required structured body, the seven reason codes with one-line guidance per code, the wave-1 SQL pattern for binding a role, the EDR_AUTHZ_SHADOW_MODE knob, and a cross-reference to the three test layers that protect the contract going forward. First-deploy guidance, not incident response.

The HTTP-end role-matrix integration test (originally proposed in the plan) is dropped: the engine matrix in
TestAllow_RoleActionMatrix already pins (role, action) verdicts, and the new handler-coverage test catches handlers that forgot the gate entirely. The narrow remaining gap (handler calls HTTPGate with the wrong Action constant) is small enough to defer past v1.

## Summary by CodeRabbit

* **Documentation**
  * Added detailed authorization enforcement documentation covering configuration, role matrices, audit logging, and operational procedures including role binding management.

* **Logging & Observability**
  * Authorization enforcement posture is now logged during startup for operational visibility and troubleshooting.

* **Tests**
  * Added architectural tests to ensure authorization enforcement mechanisms are consistently applied across all operator actions.

### DIFF
--- a/docs/authz.md
+++ b/docs/authz.md
@@ -16,12 +16,13 @@ enforcement on.
 
 The boot log announces the posture explicitly:
 
-```
-INFO authz enforcement posture=ENABLED env_var=EDR_AUTHZ_SHADOW_MODE
+```text
+INFO authz enforcement posture=ENABLED mode_description=enforcing env_var=EDR_AUTHZ_SHADOW_MODE
 ```
 
-(or `posture="SHADOW (denies audited but allowed)"` when the env
-var is set). Grep `posture=ENABLED` in your log shipper to confirm
+(or `posture=SHADOW mode_description="denies audited but allowed"`
+when the env var is set). The `posture` field is a stable enum —
+filter on `posture=ENABLED` in your log shipper to confirm
 production deployments are enforcing.
 
 ## Seeded role matrix
@@ -51,7 +52,7 @@ them from accidental delete via a future admin endpoint.
 
 When the chokepoint denies, the response is:
 
-```
+```http
 HTTP/1.1 403 Forbidden
 X-Edr-Authz-Reason: <reason>
 Content-Type: application/json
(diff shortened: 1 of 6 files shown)


getvictor:
@coderabbitai full review

PR #153, opened 2026-05-13
App Control demo cut step 2: set_application_control wire + snapshot

getvictor:
Second step of the demo cut at ai/policy/demo-plan.md. Unlocks the server's ability to push a snapshot to a host even when no rules are present yet. The Phase 3 decision engine plugs into the snapshot in the next commit.

Wire shape (server/rules/api/types.go + app_control_wire_test.go):
- CommandTypeSetApplicationControl = "set_application_control" string constant.
- SetApplicationControlPayload struct: policy_id, policy_version, rules[]. Field tags are load-bearing and Swift-decoder-binding.
- SetApplicationControlRule per-row struct: rule_type, identifier, action, enforcement, severity, custom_msg, custom_url. Disabled rules and (when now is non-zero) expired rules are filtered out by the marshal helper so the agent + extension never see them.
- MarshalSetApplicationControlPayload helper exported so the future fan-out path (Step 5) constructs the same bytes the test pins.
- Round-trip tests pin byte-shape and JSON keys; rename here breaks every deployed agent at the same instant.

Agent command (agent/commander/commander.go +
agent/commander/commander_test.go):
- New ApplicationControlSender interface — XPC bridge to the ESF extension. Mirrors the deleted-in-Phase-1 PolicySender shape but scoped to the new typed snapshot.
- Config.ApplicationControlSender field; nil means "no extension bridge" and the executor reports the command failed.
- executeSetApplicationControl decodes the envelope (policy_id > 0, policy_version > 0), forwards the raw bytes byte-identical to the extension (no re-marshal so the wire stays stable), and reports completed with {policy_id, policy_version} so operators can confirm per-host convergence via GET /commands/{id}.
- Tests: happy path (byte-equal forward), malformed payload, invalid version, missing policy_id, no sender configured.

C bridge (agent/xpcbridge/xpc_bridge.[ch]):
- xpc_bridge_send_application_control: sends "application_control.update" XPC dictionary with "data" key carrying the raw JSON. Symmetric retain/release with the existing xpc_bridge_connect contract so concurrent Disconnect during async send stays safe.

Receiver (agent/receiver/receiver.go):
- Receiver.SendApplicationControl method calls into the C bridge with the same mu-held contract the old SendPolicy used. Empty payload rejected at the Go boundary so the C side never sees an empty pointer.

Agent wiring (agent/cmd/fleet-edr-agent/main.go):
- Reintroduce the dispatcher pattern (lock-free atomic.Pointer) that bridges the per-connect *Receiver to the long-lived commander. Between disconnect and reconnect, SendApplicationControl returns an error so the command gets reported failed and the server's next fan-out re-queues the update. runReceiverLoop / runReceiverOnce regain the dispatcher arg; startCommander regains the appControlSender arg.

Extension snapshot (extension/edr/extension/ApplicationControlStore.swift):
- New singleton ApplicationControlStore with:
  * In-memory snapshot indexed by (rule_type, identifier) — six maps, one per rule_type, so the Phase 3 precedence walk is O(1) per type. The demo cut only populates BINARY; the others are zero-cost until the corresponding identifier extractors come online.
  * Lock-free reads via OSAllocatedUnfairLock so AUTH_EXEC never blocks on the snapshot.
  * Version-monotonic apply: same-or-lower version is skipped (handles duplicate XPC deliveries cleanly).
  * Atomic persist to /var/db/com.fleetdm.edr/application-control.json via write-temp-then-rename so a crash mid-write can't leave a partial file.
  * loadFromDisk on startup; missing-file fails open and the agent's next push restores state.
- ApplicationControlRule + ApplicationControlDocument Codable structs match the Go-side JSON tags exactly (rule_type, custom_msg, etc.). A rename on either side breaks the round-trip; the Go-side test pins the keys.

Extension XPC (extension/edr/extension/XPCServer.swift):
- Handler for application_control.update message type. Reads the "data" payload and hands it to ApplicationControlStore.shared.apply.

Extension startup (extension/edr/extension/main.swift):
- Call ApplicationControlStore.shared.loadFromDisk() BEFORE ESF subscribes so a racing AUTH_EXEC after process start sees the persisted snapshot, not an empty one.

## Summary by CodeRabbit

* **New Features**
  * End-to-end application control: server→agent→extension forwarding, dispatch only while a live receiver is connected, and extension applies incoming updates.
  * Extension persists and loads policy snapshots on startup; updates are applied and saved atomically with monotonic version enforcement.

* **Bug Fixes & Validation**
  * Stronger payload validation, explicit failure reporting for malformed/missing fields, and prevention of policy regressions.

* **Tests**
  * Comprehensive unit and wire-format tests covering happy paths, failures, and round-trip marshaling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/153)

### DIFF
--- a/agent/cmd/fleet-edr-agent/main.go
+++ b/agent/cmd/fleet-edr-agent/main.go
@@ -7,10 +7,12 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"log/slog"
 	"net/http"
 	"os"
 	"os/signal"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -144,14 +146,21 @@ func run() error {
 		MaxRetries: uploaderMaxRetries,
 	}, httpClient, logger)
 
+	// appControlDispatcher bridges the commander (which wants a stable
+	// ApplicationControlSender across receiver reconnects) and runReceiverLoop
+	// (which creates a new *receiver.Receiver on every connect). The ESF
+	// receiver loop publishes into this dispatcher on connect and clears it on
+	// disconnect.
+	esfAppControlDispatcher := &appControlDispatcher{}
+
 	pidTable := proctable.New()
-	go runReceiverLoop(ctx, logger, cfg.XPCService, q, pidTable, true)
+	go runReceiverLoop(ctx, logger, cfg.XPCService, q, pidTable, true, esfAppControlDispatcher)
 	if cfg.NetXPCService != "" {
-		go runReceiverLoop(ctx, logger, cfg.NetXPCService, q, pidTable, false)
+		go runReceiverLoop(ctx, logger, cfg.NetXPCService, q, pidTable, false, nil)
 	}
 	go runUploader(ctx, up, logger)
 
-	startCommander(ctx, hostID, cfg.ServerURL, tokenProvider, agentTransport, logger)
+	startCommander(ctx, hostID, cfg.ServerURL, tokenProvider, esfAppControlDispatcher, agentTransport, logger)
 	go pruneLoop(ctx, q, cfg.PruneAge, logger)
 	startProcessReconciler(ctx, cfg, pidTable, q, tokenProvider, logger)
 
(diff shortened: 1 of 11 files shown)


getvictor:
@coderabbitai full review

PR #154, opened 2026-05-13
App Control demo cut step 3: AUTH_EXEC BINARY decision + sha256 cache

getvictor:
Third step of the demo cut at ai/policy/demo-plan.md. The extension now actually blocks execs against the snapshot. Demo beat #6 (modal pop-up after a denied exec) is reachable end-to-end pending the host- app modal that Step 6 adds.

FileHashCache (extension/edr/extension/FileHashCache.swift):
- New singleton FileHashCache. SHA-256 hex strings indexed by (inode, mtimeSec, mtimeNsec) tuple from es_file_t.stat. Keying on the inode triple rather than the path catches "same file, different path" cases (hard links, symlinks resolved upstream) that a path key would miss.
- Three call paths:
  * lookup(stat:) — pure read, used on AUTH (deadline-sensitive).
  * lookupOrCompute(path:stat:) — sync compute, used on NOTIFY (no deadline). Every event carries a real hash when the file is readable.
  * startLazyFill(path:stat:) — async compute on AUTH cache miss. Recheck-under-lock guard avoids double-compute when a concurrent sync compute races the async dispatch.
- Streaming SHA-256 via CryptoKit's incremental SHA256 over 64 KiB FileHandle chunks so multi-gigabyte binaries don't have to fit in memory.
- OSAllocatedUnfairLock around the dictionary; concurrent compute queue (qos: .userInitiated) so async fills can run in parallel across distinct binaries.

AUTH_EXEC decision walk (extension/edr/extension/ESFSubscriber.swift):
- handleAuthExec now consults the ApplicationControlStore snapshot:
  1. Failsafe — exec target's team_id matches extensionTeamID ("FDG8Q7N4CC") → ALLOW unconditionally. Prevents a misconfigured rule from blocking the agent / extension / host app. Hardcoded for the demo; the Phase B server-pushed failsafe list widens this to launchd, systemextensionsd, and admin carve-outs.
  2. FileHashCache.lookup(stat:) cache miss → ALLOW + startLazyFill. Documented "first exec misses, second exec catches" behavior the demo recording exercises (warm the cache once, then push the rule, then re-launch — the second launch hits the block).
  3. Cache hit + BINARY rule with action=BLOCK + enforcement=PROTECT → ES_AUTH_RESULT_DENY. Logged at WARN with the matched hash so post-mortem inspection in `log show` is straightforward.
  4. Otherwise → ALLOW.
- Decision is reached inside the AUTH_EXEC deadline because every step is a constant-time map lookup; file read + signing-info fetch happen off the callback on the lazy-fill background queue.

Exec event hash population (handleExec):
- The existing `sha256` field on ExecPayload was previously hard- coded to nil. Now reads from FileHashCache.lookupOrCompute on every NOTIFY_EXEC. Common case is a cache hit because AUTH preceded NOTIFY for the same exec and triggered the lazy fill; cold cache computes synchronously on this path (no deadline) so every event emitted carries a real hash for the alerts pipeline + UI.
- nil is still possible if the file disappears between AUTH and NOTIFY (rare race); the event still fires with the rest of the metadata rather than dropping entirely.

Startup log (start()):
- Switched the "Application Control disabled (phase 1)" warning to an info-level "Application Control active" line so operators monitoring logs during the demo see the right state.

Demo readiness:
- Beat #5 (launch Calculator on edr-dev) now produces a real AUTH_EXEC → snapshot lookup → DENY when the SHA-256 is in the snapshot. Beat #6 (modal pop-up) still depends on Step 6.
- Beat #8 (alert in the UI) depends on Step 4 wiring the block event through the detection pipeline.

## Summary by CodeRabbit

* **New Features**
  * Added application control rules with action types (block, allow, silent block) and enforcement modes (protect, detect)
  * Implemented executable hash-based verification for enhanced security control enforcement with caching to optimize performance
  * Updated application control decision logic to evaluate cached file signatures against configured rules

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/154)

### DIFF
--- a/extension/edr/extension/ApplicationControlStore.swift
+++ b/extension/edr/extension/ApplicationControlStore.swift
@@ -81,6 +81,22 @@ enum ApplicationControlRuleType {
     static let path = "PATH"
 }
 
+/// Rule-action tokens that match the server enum exactly. The demo cut
+/// only emits BLOCK; ALLOW and SILENT_BLOCK arrive with Phase B Lockdown.
+enum ApplicationControlAction {
+    static let block = "BLOCK"
+    static let allow = "ALLOW"
+    static let silentBlock = "SILENT_BLOCK"
+}
+
+/// Rule-enforcement tokens that match the server enum exactly. The demo
+/// cut only enforces PROTECT (deny-on-match); DETECT (log only) arrives
+/// with Phase B.
+enum ApplicationControlEnforcement {
+    static let protect = "PROTECT"
+    static let detect = "DETECT"
+}
+
 /// ApplicationControlStore holds the typed in-memory snapshot consulted by
 /// the AUTH_EXEC decision engine (Step 3). The snapshot also persists to
 /// disk so the extension's policy survives restarts. Updates from the agent
--- a/extension/edr/extension/ESFStringToken.swift
+++ b/extension/edr/extension/ESFStringToken.swift
@@ -8,12 +8,20 @@ import Foundation
 /// overread into adjacent memory in the worst case and produce garbage
 /// in the typical case. Use this helper everywhere a Swift String is
 /// derived from an es_string_token_t — paths, team IDs, signing IDs.
+///
 /// Returns the empty string when the token has no payload (data is
-/// nil or length is zero).
+/// nil or length is zero) OR when the bytes are not valid UTF-8. The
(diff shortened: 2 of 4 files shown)


getvictor:
@coderabbitai full review

PR #155, opened 2026-05-14
App Control demo cut step 4: block event -> alert (unified detection …

getvictor:
…view)

Closes demo beat #8: an AUTH_EXEC denial on the host produces a row in the alerts view alongside catalog-rule findings.

Server side
- detection/api/types.go: new AlertSourceDetection / AlertSourceApplicationControl constants; Source field on Alert, Finding, and AlertFilter.
- detection/bootstrap/schema.go: source ENUM('detection', 'application_control') column on alerts (default 'detection'); the unique dedup key promotes to (source, host_id, rule_id, process_id) so a catalog rule and an app-control rule that happen to share an identifier never collapse into one row.
- detection/internal/mysql/alerts.go: source threaded through every insert / select / count / dedup lookup; defense-in-depth defaults blank Source to 'detection' so callers that forget the field land in the catalog-rule bucket instead of hitting MySQL's truncated- ENUM error.
- detection/internal/engine/engine.go: persistFinding stamps the Finding's Source onto the alert, defaulting to AlertSourceDetection when blank (catalog rules don't need to set it).

Rule
- rules/internal/catalog/application_control_block.go: built-in catalog rule that maps every application_control_block ingest event to a Finding with Source=application_control. Description prefers the operator's custom_msg, falling back to "Blocked <rule_type> rule for <identifier>" per spec.
- rules/internal/catalog/registry.go: register the rule alongside the catalog detection rules so it loads via the existing rulesCtx.ContentService() -> detectionCtx.LoadActive() path.

Wire
- rules/api/types.go: new RuleID string field on SetApplicationControlRule + ApplicationControlRuleID() helper that renders "app_control:<numeric_id>" so the extension echoes a stable string identifier in the block event; alert dedup never collides with catalog rule ids.
- schema/events.json: register application_control_block event_type and the matching payload schema.

Extension
- ApplicationControlStore.swift: ruleID on the Codable rule struct; decodes the new rule_id wire field.
- EventSerializer.swift: ApplicationControlBlockPayload Codable matching the Go decode struct field-for-field.
- ESFSubscriber.swift: emitBlockEvent() builds the payload and hands it to onEvent after the DENY response so the JSON encode + XPC handoff happen off the AUTH callback's kernel deadline.

UI
- types.ts: Alert.source field so the alerts page can chip / filter by origin without re-parsing rule_id (the filter chip itself is step 9 polish; this lands the field so step-5 / step-9 don't have to retroactively widen the type).

Tests
- rules/internal/catalog/application_control_block_test.go: table- driven unit test for every branch of the rule (happy path, missing rule_id / severity, missing process row, custom_msg default), plus the GraphReader-error propagation case.
- rules/internal/catalog/registry_test.go + rules/internal/tests/ integration_test.go: count + roster updated to include the new rule.
- rules/api/app_control_wire_test.go: rule_id round-trip pinned in both the example-based and rapid PBT round-trip tests.
- test/integration/app_control_block_test.go: cross-context test exercising the demo path end-to-end -- agent posts a block event, it surfaces as an alert with source=application_control; re-post the same block dedups into one row; missing custom_msg lands the spec's deterministic default description.

## Summary by CodeRabbit

* **New Features**
  * Application control blocking now generates tracked alerts when execution attempts are denied
  * Alerts include a source field to distinguish between detection-based and application control-based origins
  * Application control block events support optional custom messages and URLs for operator context

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/155)

### DIFF
--- a/extension/edr/extension/ApplicationControlStore.swift
+++ b/extension/edr/extension/ApplicationControlStore.swift
@@ -36,7 +36,12 @@ struct ApplicationControlSnapshot {
 /// server's `server/rules/api.SetApplicationControlRule` JSON tags — a
 /// rename on either side is a contract break and the round-trip test on the
 /// server side is the gate.
+///
+/// ruleID is the stable string identifier (e.g. "app_control:42") the
+/// AUTH_EXEC handler echoes back in the `application_control_block` event
+/// so the server's alert mapping lands the alert under the same rule_id.
 struct ApplicationControlRule: Codable {
+    let ruleID: String
     let ruleType: String
     let identifier: String
     let action: String
@@ -46,6 +51,7 @@ struct ApplicationControlRule: Codable {
     let customURL: String?
 
     enum CodingKeys: String, CodingKey {
+        case ruleID = "rule_id"
         case ruleType = "rule_type"
         case identifier
         case action
--- a/extension/edr/extension/ESFSubscriber.swift
+++ b/extension/edr/extension/ESFSubscriber.swift
@@ -126,7 +126,16 @@ final class ESFSubscriber: Sendable {
                rule.action == ApplicationControlAction.block,
                rule.enforcement == ApplicationControlEnforcement.protect {
                 logger.warning("AUTH_EXEC DENIED by application control: \(path, privacy: .public) sha256=\(sha256, privacy: .public)")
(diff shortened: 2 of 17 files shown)


getvictor:
@coderabbitai full review

PR #157, opened 2026-05-14
App Control demo cut step 6: host-app notification modal (Phase B pul…

getvictor:
…led forward)

Closes demo beat #6: a denied AUTH_EXEC pops a Santa-shaped NSAlert on the user's desktop with the operator's custom_msg and an optional "More info" link.

Wire shape
- New XPC channel: extension → host app via Mach service `FDG8Q7N4CC.com.fleetdm.edr.notifications`. Message type `application_control.block_notification`; payload is a JSON-encoded BlockNotificationPayload (Codable on both sides, mirrored in extension/edr/extension/BlockNotification.swift and extension/edr/edr/BlockNotification.swift). Code-signing peer requirement matches the agent ↔ extension channel: any binary signed by team FDG8Q7N4CC.

Extension side
- NotificationClient.swift: lazy outbound XPC connection with the standard team-id peer-requirement gate. Serial queue keeps the ordering predictable; xpc_connection_send_message is fire-and-forget so a missing host app never stalls the AUTH_EXEC handler. Reconnects on the next notify when the cached connection invalidates.
- ESFSubscriber.swift: after the existing DENY response + audit event emission, fire `NotificationClient.shared.notify(payload)`. Runs after the kernel is already unblocked so the alert is post-hoc UX, not authorization.

Host-app side
- NotificationListener.swift: XPC listener mirroring the extension's XPCServer shape (same listener + per-peer event handler + peer-validation pattern). Decodes Codable payload and forwards to a BlockAlertPresenter.
- BlockAlert.swift: BlockAlertPresenterAppKit renders one NSAlert per accepted notification. Serial DispatchQueue + main-thread hop so two back-to-back denials don't stack overlapping modals; (rule_id, binary_path) tuples seen within a 30s dedup window are suppressed so a misbehaving caller can't paper the screen. "Dismiss" closes the modal; "More info" only renders when custom_url is set + scheme is http/https, and opens it via NSWorkspace.shared.open.
- main.swift: new `notify` mode that boots NSApplication with accessory activation policy (no Dock icon), starts the listener, and runs the AppKit run loop. Distinct from the other CLI modes' dispatchMain() because NSAlert needs the AppKit run loop's event dispatch.
- Info.plist: MachServices declaration so launchd registers the service name when the host app is run via the new LaunchAgent.

Deployment
- com.fleetdm.edr.notify.plist: per-user LaunchAgent that runs the host app in `notify` mode. KeepAlive=false + on-demand Mach-service binding so a crashed presenter doesn't crashloop on the user's desktop; the next AUTH_EXEC denial spawns a fresh instance.

Demo dry-run path (per ai/policy/demo-plan.md beat #6): on the SIP-disabled VM, install the LaunchAgent OR start the surface manually with `sudo /Applications/Fleet\ EDR.app/Contents/MacOS/edr notify` in a Terminal session. The next blocked exec pops the modal.

Out of scope (Phase B): hardening the modal into a true authorization dialog, custom-styled NSWindow, per-policy mute, notification-center integration.

### DIFF
--- a/extension/edr/com.fleetdm.edr.notify.plist
+++ b/extension/edr/com.fleetdm.edr.notify.plist
@@ -51,9 +51,24 @@
 	<key>ProcessType</key>
 	<string>Interactive</string>
 
-	<key>StandardOutPath</key>
-	<string>/tmp/fleet-edr-notify.log</string>
-	<key>StandardErrorPath</key>
-	<string>/tmp/fleet-edr-notify.log</string>
+	<!--
+	    No StandardOutPath / StandardErrorPath: the host app routes
+	    its output through os_log (Logger(subsystem: "com.fleetdm.edr"))
+	    so anything worth keeping shows up in the unified log and is
+	    queryable via `log show --predicate 'subsystem == "com.fleetdm.edr"'`.
+	    A shared file path under /tmp would race across users on a
+	    multi-user install (the first user's launchd creates the file
+	    with permissions that lock out later sessions) — caught by
+	    Gemini + Copilot on PR #157.
+
+	    Packaging note: the dist/pkg path does NOT yet copy this
+	    plist into /Library/LaunchAgents. The demo dry-run runs the
+	    host app manually in a Terminal session
+	    (`sudo /Applications/Fleet\ EDR.app/Contents/MacOS/edr notify`)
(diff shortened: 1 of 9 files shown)


getvictor:
@coderabbitai full review

PR #158, opened 2026-05-14
App Control demo cut step 7: UI for the Application Control admin sur…

getvictor:
…face

Unlocks demo beats #1, #2, and #3 — the camera-facing admin workflow (open Application Control, navigate to Default policy, paste a SHA-256, save).

Routes + nav
- New top-nav entry "Application control" between "Alerts" and "Coverage" (matches the demo plan's section F).
- New routes mounted via a feature-local ApplicationControlRoutes router so App.tsx stays shallow:
    /app-control                  → PoliciesList
    /app-control/policies/:id     → PolicyDetail

API clients
- types.ts: ApplicationControlPolicy + ApplicationControlRule interfaces mirror the server wire shape.
- api.ts: listAppControlPolicies, getAppControlPolicy, and createAppControlRule. Typed AppControlApiError surfaces the handler's `application_control.*` error codes so callers can switch on the code (e.g. duplicate_rule → inline "this identifier already exists" hint) rather than parse the message. POST routes through the existing CSRF / reauth challenge path for parity with the other state-changing endpoints.

Pages
- PoliciesList: renders the seed Default with rule count, version, last modified, and the hardcoded "all hosts" assignment. The "New policy" button renders disabled with the demo-cut "coming soon" tooltip (multi-policy is post-demo per plan section F).
- PolicyDetail: header + back link, rules table with the columns the demo plan calls for (type, identifier truncated to 16 hex chars with full value on hover, severity badge, custom message, last modified). Per-row Edit / Disable / Delete render disabled with the same coming-soon tooltip pattern. Refresh-on- save via a refreshKey counter so the new rule shows up without a manual reload.

Add Rule modal
- Native <dialog> opened via showModal() so the browser provides the backdrop, focus trap, and Escape-to-cancel — same pattern ReauthModal already uses.
- Type selector lists every rule_type value the schema's ENUM accepts, but only BINARY is selectable; the rest are rendered disabled with a "coming soon" suffix so the camera sees the roadmap without faking behaviour.
- Client-side validator mirrors the server's BINARY shape (64 lowercase hex chars) so the operator gets immediate feedback instead of a round-trip + a typed 400.
- Reason field is required (gates Save) so every authored rule has the audit-log row's `reason` populated.
- AppControlApiError → human-readable copy via an explicit Map; unrecognised codes fall back to the server's `message` field rather than disappearing.
- Reauth-gated via useReauthRetry so a critical-severity rule creation past the freshness window prompts the operator through the existing flow.

Out of scope for the demo cut (still in the OpenSpec backlog):
- Paste-many flow (8.5 in tasks.md)
- React Testing Library coverage (8.7) — the demo plan calls for visual polish iteration on the dry-run rather than upfront RTL tests; will land alongside the post-demo CRUD endpoints when the surfaces stabilise.

### DIFF
--- a/server/cmd/fleet-edr-server/main.go
+++ b/server/cmd/fleet-edr-server/main.go
@@ -555,6 +555,16 @@ func registerSessionRoutes(mux *http.ServeMux, d muxDeps) {
 		"GET /api/rules",
 		"GET /api/audit-events",
 		"GET /api/session",
+		// Application Control demo cut. rulesCtx.RegisterAuthedRoutes
+		// mounts these on apiMux; the outer router needs each path
+		// enumerated here so requests reach the session-protected
+		// wrapper instead of falling through to the `/` catchall and
+		// 302 → /ui/. Surfaced by the step-8 dry-run on PR #158 — the
+		// list-policies API returned the SPA's index.html, which the
+		// UI fetch parsed as JSON and errored with "Unexpected token '<'".
+		"GET /api/v1/app-control/policies",
+		"GET /api/v1/app-control/policies/{id}",
+		"POST /api/v1/app-control/policies/{id}/rules",
 		// Break-glass reauth ceremony. The handlers are mounted on
 		// apiMux via identityCtx.RegisterAuthedRoutes, but the outer
 		// router needs each path enumerated here so the session-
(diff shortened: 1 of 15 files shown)


getvictor:
@coderabbitai full review

PR #163, opened 2026-05-15
dev: live-dir UI bundle + ADR gap-list entries

### LINKED ISSUE #141: Dev experience: serve UI from disk in dev so task build:ui refreshes without restarting dev:server
## Problem

`server/ui/embed.go` uses `//go:embed all:dist` to embed the React bundle at compile time. In production this is correct - the binary ships with the UI baked in. In dev it creates a paper cut:

1. Run `task dev:server` - server compiles, embeds whatever is currently in `server/ui/dist/`, starts serving.
2. Run `task build:ui` - new bundle written to `server/ui/dist/` with new hashed asset names.
3. Browser hard-reload - server still serves the **old** bundle because its embedded FS is frozen at compile time. New CSS hash on disk is invisible.

The fix today is "Ctrl-C and re-run `task dev:server` after every UI rebuild," which is friction the operator has to remember and a real time sink during QA when iterating on layout fixes.

`task dev:ui` (Vite HMR via :5173 + proxy to :8088) sidesteps this entirely and is the right tool for active UI iteration. But for QA against the production embedding shape, the operator still needs the bundled-server view, and that view doesn't refresh.

## Proposed fix

A small fallback in `server/ui/embed.go` (or its caller in the detection intake handler / cmd/main): when an env flag is set, read from `os.DirFS("server/ui/dist")` instead of the embedded FS. Vite writes new files atomically on each build; the server picks them up on the next HTTP request without a restart.

```go
// server/ui/embed.go
package ui

import (
    "embed"
    "io/fs"
    "os"
)

//go:embed all:dist
var embeddedDist embed.FS

// FS returns the filesystem the server should serve UI assets from.
// Production: the embedded bundle (immutable, ships with the binary).
// Dev: when EDR_UI_LIVE_DIR is set, reads from disk so `task build:ui`
// refreshes the bundle without a `task dev:server` restart.
func FS() fs.FS {
    if dir := os.Getenv("EDR_UI_LIVE_DIR"); dir != "" {
        return os.DirFS(dir)
    }
    sub, _ := fs.Sub(embeddedDist, "dist")
    return sub
}
```

Wire `Taskfile.yml`'s `dev:server` task to set `EDR_UI_LIVE_DIR=server/ui/dist` so the dev workflow picks it up automatically. Production builds set nothing -> embedded path -> identical to today.

## Scope

- `server/ui/embed.go`: small refactor to expose `FS()` instead of (or alongside) the bare `DistFS`.
- Caller(s) of `DistFS`: switch to `FS()`. Probably one or two sites.
- `Taskfile.yml`: add `EDR_UI_LIVE_DIR: server/ui/dist` to `dev:server`'s env block.
- New unit test confirming `FS()` returns the embedded copy when the env var is unset and the live path when it's set.
- Doc update in `docs/operations.md` "Quick reference" or a new "Dev workflows" section noting the env knob exists.

## Acceptance

- `task dev:server` running, with no source changes, behaves identically to today (same hashes served, same headers).
- `task dev:server` running, then `task build:ui` (in another terminal), then a browser reload -> the new bundle is served without restarting the Go process.
- Production binary build (`task build:server`) does NOT change behaviour - embedded FS is the only path because the env var is unset.

## Out of scope

- Live-reload of the Go server itself (no built-in HMR for compiled Go; out of scope and well-served by tools like `air` if anyone wants it).
- Watch mode for the UI build (use `task dev:ui` for HMR; this issue is about the embedded-bundle flow, not the Vite-served flow).

## Priority

LOW - workaround is "restart `task dev:server` after `task build:ui`." Fix is small and removes a recurring friction during UI QA passes against the bundled-server shape.

## Cross-references

Surfaced during the user-management QA pass while iterating on auth-page layout fixes (commit XXXX). The QA plan at `claude/qa/user-management.md` documents the current "rebuild + restart" workaround in its Option 1 setup block; once this issue lands, the QA plan can drop the rebuild step.

### LINKED ISSUE #145: ADR audit gap: document MySQL-only and MDM-only-update decisions
## Context

Surfaced by the 2026-Q2 best-practices refresh (see `docs/maintenance/log.md`,
2026-05-12). `docs/best-practices.md` carries two `[-] will not do` items
whose rationale is load-bearing architecture, not stylistic preference, and
which currently live only in the best-practices doc rather than in an ADR.
Per `docs/maintenance/tasks/adr-audit.md`, that's the gap-list trigger.

## Candidates for new ADRs

1. **MySQL-only data plane (no PostgreSQL)**
   - Load-bearing because every store-layer choice (DSN, migrations,
     `FOR UPDATE SKIP LOCKED`, parseTime quirks, the in-process idempotent
     ALTER pattern) is justified by it.
   - Trade-off captured in best-practices.md §10: supporting two RDBMSes
     doubles migration testing, query tuning, and store-layer surface for a
     small team; the customer escape hatch is Debezium CDC from MySQL.

2. **In-product auto-update path: MDM-only by design**
   - Load-bearing because it shapes the installer contract, the lack of a
     Sparkle-style update channel, the lack of a manifest-fetcher in the
     agent, and the explicit dependence on Fleet / Jamf / Kandji / Intune
     for the update channel.
   - Trade-off captured in best-practices.md §11: in-product self-update
     bypasses customer change management and gets flagged by procurement;
     Sparkle is a consumer-app pattern.

Both decisions are currently single-sentence rationales next to a checkbox.
Both reverse-engineerable from current code, but only with effort.

## Existing ADR-audit gap list

`docs/maintenance/tasks/adr-audit.md` already tracks several gaps
(OTel-only metrics, test layering, embedded UI, the `testdb/full.Open`
seam, `Co-Authored-By` policy, AI-tooling-as-code). This issue adds the
two above to that gap list rather than writing the ADRs in-line.

## Out of scope

Writing the ADRs themselves; that needs its own focused session per the
adr-audit task guidance ("don't write the new ADR during this audit").

getvictor:
Fixes #141: EDR_UI_LIVE_DIR opts the server into reading server/ui/dist from disk instead of the compile-time embedded copy, so a task build:ui in another terminal refreshes the bundle without restarting dev:server. Production binaries leave the env unset and serve the embedded path; the existing fs.Sub error branch in registerUIRoutes is preserved.

Fixes #145: adds MySQL-only data plane and MDM-only auto-update rows to the ADR audit gap-check table so the next audit pass picks them up.

## Summary by CodeRabbit

* **New Features**
  * Development workflow improvement: UI changes are now automatically picked up on the next request without requiring a server restart.

* **Documentation**
  * Updated architecture documentation with development and production UI serving behavior.
  * Enhanced task audit documentation with additional decision domain coverage.

* **Tests**
  * Added test coverage for UI bundle loading in development and production modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/163)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -360,6 +360,11 @@ tasks:
       EDR_HOST_TOKEN_GRACE 5m. Override either to a short value (e.g.
       EDR_HOST_TOKEN_LIFETIME=30s) when QA-ing the rotation flow against a real VM agent
       so you can observe a rotation cycle inside one minute rather than next-day.
+
+      EDR_UI_LIVE_DIR (issue #141) points the server at server/ui/dist on disk instead of
+      the compile-time embedded bundle, so `task build:ui` is picked up on the next
+      request without restarting the server. Production binaries leave the var unset and
+      serve the embedded copy.
     env:
       EDR_DSN: "root:@tcp(127.0.0.1:3316)/edr?parseTime=true"
       EDR_ENROLL_SECRET: dev-enroll-secret
@@ -368,6 +373,7 @@ tasks:
       EDR_LOG_FORMAT: text
       EDR_HOST_TOKEN_LIFETIME: 24h
       EDR_HOST_TOKEN_GRACE: 5m
+      EDR_UI_LIVE_DIR: server/ui/dist
       # Dev-only deterministic session HMAC key. Required when break-glass or OIDC is
       # configured (the server refuses to boot without it). Production deployments
       # MUST supply a real random secret via EDR_SESSION_SIGNING_KEY_FILE (Docker
@@ -402,12 +408,12 @@ tasks:
   dev:
     desc: |
       Run backend + UI dev servers in parallel for fast iteration. Backend lives at
-      http://127.0.0.1:8088 (Go server, serves the bundled `server/ui/dist/` -- so it
-      shows the LAST `task build:ui` output, not your live edits); UI HMR lives at
-      http://localhost:5173/ui/ (Vite dev server, proxies /api/* back to :8088). Edit
-      .tsx/.scss -> browser updates in <1s; edit .go -> Ctrl-C and re-run `task dev`
-      (Go has no built-in HMR). For QA against the embedded bundle (production shape),
-      use `task build:ui && task dev:server` instead.
+      http://127.0.0.1:8088 (Go server, serves `server/ui/dist/` via EDR_UI_LIVE_DIR --
+      a `task build:ui` in another terminal is picked up on the next request without
+      restarting the server, per issue #141); UI HMR lives at http://localhost:5173/ui/
+      (Vite dev server, proxies /api/* back to :8088). Edit .tsx/.scss -> Vite HMR
+      updates in <1s; edit .go -> Ctrl-C and re-run `task dev` (Go has no built-in HMR).
+      For QA against the embedded bundle (production shape), unset EDR_UI_LIVE_DIR.
     deps:
       - dev:server
       - dev:ui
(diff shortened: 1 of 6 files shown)


getvictor:
@coderabbitai full review

Issue #68, opened 2026-04-29
Expand how executables can be blocked (like Santa)


getvictor:

getvictor:
Closing as superseded.

The openspec scope of this umbrella shipped in PR #184 (demo-cut task annotation + Detect-mode change draft). The
remaining concrete work, expanding rule-type enforcement beyond BINARY, is re-scoped to a fresh issue so each
follow-on PR has a clear, single-purpose target. CERTIFICATE in particular is broken out as the highest-leverage
incident-response rule type per the third-party Santa-comparison review.

Successor: app control rule-type expansion (CDHASH / SIGNINGID / TEAMID / CERTIFICATE / PATH).

Per-rule Detect mode is tracked separately via the `add-application-control-detect-mode` openspec change that #184
drafted; ALLOW and SILENT_BLOCK actions, multi-policy host groups, and CERTIFICATE/PATH on the action enum are
v0.2.0 / Phase B.

PR #197, opened 2026-05-19
app-control: per-policy assignment_count + overlap fan-out dedup test…

getvictor:
… (#68 / 11.1.5 + 11.1.6)

Closes the last code-side items in the Phase A close-out (PR-1) for #68:

11.1.6 — PoliciesList real assignment count
  Add AssignmentCount (json:"assignment_count") to api.ApplicationControlPolicy.
  Server-decorate via a shared policySelectWithAssignmentCount LEFT JOIN against
  an app_control_assignments aggregate sub-query; COALESCE turns the LEFT-JOIN NULL
  for policies with zero assignments into the integer zero scan target expects.
  Three policy fetches share the JOIN: GetPolicyByName, getPolicyByID, ListPolicies.
  A shared scanPolicyRow helper pins the column order across all three so a future
  SELECT-list change surfaces as a typed scan error rather than a silent field swap.

  UI: PoliciesList renders the value via an assignmentLabel helper -- singular
  ("1 host group") at exactly count==1, plural at 2+, and an explicit "no host
  groups" at zero (an admin posture, not a wire-shape error). The hardcoded
  "all hosts" placeholder is gone; a regression that wires it back in is caught
  by the existing test asserting that text doesn't render.

11.1.5 — Overlap fan-out dedup test
  New TestAppControlREST_CreateRule_FanOutDedupsAcrossOverlappingAssignments inserts
  a second host_group with {"type":"all"} criteria + a second assignment row to the
  Default policy directly via SQL (Phase A's REST surface gates non-all-hosts group
  mutations behind 405). A single rule POST then exercises service.fanout's seen-set
  union: each of N hosts receives exactly ONE set_application_control command even
  though it is reachable via two assignments, and the audit row's fanout_hosts
  reflects unique hosts not assignment-fan crosses.

  Without this test, a regression that drops the seen-set or skips the across-groups
  union still passes the existing dedup test (which only exercises HostLister-side
  duplicates inside one group resolve, not the across-assignments union).

Local verification
  - task lint:go (custom golangci-lint with commentwrap): 0 issues across ./server/rules/... and ./test/integration.
  - go test ./server/rules/internal/appcontrol/... + the bootstrap + catalog
    + testdb suites: all pass.
  - go test -tags integration -count=1 ./server/rules/internal/tests/ and ./test/integration/: all pass against the local MySQL test instance, including the new overlap dedup case.
  - npm run build (ui): tsc + vite build clean.
  - npx vitest run on PoliciesList + PolicyDetail: 22/22 tests pass.

The remaining Phase A items are documentation-only (11.5.1: CLAUDE.md note, 11.5.2: developer-facing docs/ page). Those land as the final PR before the change can be archived.

## Summary by CodeRabbit

* **New Features**
  * ApplicationControl policies now display the number of host groups they are assigned to in the policies list.
  * Assignment counts include proper singular/plural formatting ("1 host group", "3 host groups", "no host groups").
  * Removed previous hardcoded "all hosts" placeholder text.

* **Tests**
  * Added integration test validating fan-out deduplication across overlapping policy assignments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/197?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/detection/internal/tests/integration_test.go
+++ b/server/detection/internal/tests/integration_test.go
@@ -708,10 +708,18 @@ func TestGraph_ExecPayloadCDHashRoundTrips(t *testing.T) {
 			Payload: json.RawMessage(`{"pid":1002,"ppid":1,"path":"/usr/bin/legacy","args":["legacy"],"sha256":"cafef00d"}`)},
 	})
 
+	// Wait until BOTH rows have landed before asserting either one's shape. The processor is async and the two exec events
+	// are not guaranteed to materialise in the order they were posted; without this combined wait, a CI run that races the
+	// 1001 Eventually-pass against the 1002 row commit would return nil for nonHR even though both are inflight (issue
+	// caught in the PR #197 CI run where 1001 landed first and 1002 was still queued when the test asserted).
 	require.Eventually(t, func() bool {
-		p, err := d.Service().GetProcessDetail(ctx, "h-cdh", 1001, now+1)
-		return err == nil && p != nil && p.Process.CDHash != nil && *p.Process.CDHash == cdhash
-	}, 5*time.Second, 50*time.Millisecond, "HR exec must persist cdhash on the processes row")
+		hr, err := d.Service().GetProcessDetail(ctx, "h-cdh", 1001, now+1)
+		if err != nil || hr == nil || hr.Process.CDHash == nil || *hr.Process.CDHash != cdhash {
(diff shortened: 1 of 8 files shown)


getvictor:
@coderabbitai full review

PR #206, opened 2026-05-20
M1: split receiver behind build tags so agent compiles on linux

getvictor:
First milestone of the UAT plan in docs/testing-strategy.md. The Go agent's core (queue, uploader, commander, reconciler, etc.) is already platform-portable; only agent/receiver pulls in macOS-only XPC via CGo (#cgo LDFLAGS: -framework Foundation, #include <xpc/xpc.h>). That single package was making `GOOS=linux go build ./agent/...` fail with "build constraints exclude all Go files in agent/receiver", which in turn blocks the headless-agent + server integration job (M3) and any future Linux-side developer ergonomics.

This change:

- Tags the existing CGo files //go:build darwin (receiver.go, callbacks.go) so they are explicit about their platform, not implicit via the cgo cross-compile default of CGO_ENABLED=0.
- Extracts platform-neutral identifiers (Event struct, Error* constants, SetLogger + logger atomic) into a new untagged common.go so the linux build sees them without duplicating definitions.
- Adds receiver_other.go (//go:build !darwin) with a stub Receiver that mirrors the darwin Receiver's public surface (New / Events / Errors / Connect / SendApplicationControl / Ping / Disconnect). Connect, SendApplicationControl, and Ping return a sentinel ErrUnsupported so the agent fails closed on linux today; the inject-able variant for tests is M2's scope.
- Adds task build:agent:linux as the named acceptance check and wires it into the agent-test CI job (runs on macos-latest via cross-compile) so the platform-portability invariant stays green going forward.

Verified locally: task build:agent (darwin), task build:agent:linux, task test:go:agent, task lint:go all pass. xpcbridge needs no changes because it has no .go files; only the darwin receiver references its C headers. hostid is already pure-Go and compiles on linux today.

No production darwin behaviour change. Refs UAT plan M1 milestone.

## Summary by CodeRabbit

* **Refactor**
  * Refactored internal architecture to centralize shared logging definitions, event handling, and error codes across platform implementations. The codebase now provides native support for macOS with compatibility stubs for other platforms.

* **Chores**
  * Enhanced CI test workflow with a new cross-compilation validation step to ensure the agent builds correctly for Linux platforms, improving portability and compatibility checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/206?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,6 +122,13 @@ jobs:
           version: 3.50.0
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Cross-build agent for linux (UAT plan M1 invariant)
+        # The agent's Go core must compile on linux so the headless integration
+        # job (UAT plan M3) and developer Linux machines can build it. The
+        # darwin-only XPC bridge is selected by build tags; everything else is
+        # platform-portable. See docs/testing-strategy.md.
+        run: task build:agent:linux
+
       - name: Run agent tests with coverage
         run: task test:go:agent:coverage
 
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -50,6 +50,20 @@ tasks:
     cmds:
       - go build -o {{.AGENT_BIN}} {{.AGENT_PKG}}
 
+  build:agent:linux:
+    desc: |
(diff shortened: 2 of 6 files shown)


getvictor:
@coderabbitai full review

PR #213, opened 2026-05-20
M4: L3 headless-agent + server integration test

getvictor:
Fourth milestone of the UAT plan in docs/testing-strategy.md. With M1's build-tag split, M2's headless binary, and M3's fakeagent library all merged, M4 closes the L3 loop: a Go integration test that boots the real EDR server (via test/integration.Setup), runs the real headless agent in-process, drives the M3 starter scenario corpus through the agent's POST /event control plane, and asserts events landed in the server's detection service. End-to-end proof that scenario YAML -> fakeagent envelopes -> headless control plane -> stub receiver -> queue -> uploader -> /api/events -> processor -> events table works on every PR.

What this lands:

- agent/cmd/fleet-edr-agent-headless/headless/ is a new importable sub-package containing Run, Options, the receiver-pump goroutine, the control-plane HTTP handlers, and the M2 unit tests, all moved from the cmd directory's main package. main.go shrinks to a thin entrypoint that imports the new package. The split exists so the L3 integration test can call headless.Run in-process without building a subprocess. Production behaviour is byte-identical; the build tag (`//go:build !darwin || !cgo`) carries over.

- test/integration/agentserver/agentserver_test.go is the new L3 test. One subtest per scenario in the M3 corpus (5 total after this PR). Each subtest: enrolls a fresh host against the Setup Stack's /api/enroll, builds headless.Options pointing at the Stack URL with a stub TokenProvider carrying the minted token, spins headless.Run in a goroutine, waits for the control plane to bind, feeds the scenario via fakeagent.FeedControlPlane, then Eventually-polls detection.Service.ListHosts until EventCount matches len(timeline). Build tag: `integration && (!darwin || !cgo)` so CI ubuntu-latest picks it up automatically and darwin devs opt-in via CGO_ENABLED=0.

  Subtests run SERIALLY rather than t.Parallel because each Setup Stack opens its own MySQL pool; five concurrent Stacks on top of the rest of the integration suite's parallel load consistently push past MySQL's max_connections ceiling. Sequential keeps the package's peak connection count to ~4 and wall time stays under 3 seconds.

- test/fakeagent/scenarios/ gains 2 new YAML files so the M4 corpus = 5 total: process-tree-deep.yaml (4-level parent->child process chain, exercises depth) and mixed-events.yaml (fork+exec+ open+dns+network+exit covering 5 event types in a single timeline).

[...]

PR #212, opened 2026-05-20
M3: fakeagent library for scenario-driven event injection

getvictor:
Third milestone of the UAT plan in docs/testing-strategy.md. With M1's build-tag split letting the agent compile on linux and M2's headless binary exposing a unix-socket POST /event control plane, M3 ships the library that consumers (the M4 L3 integration test, the M10 detection- efficacy corpus, Playwright fixtures, the M12 scale test) use to drive the headless binary or the server directly from YAML scenarios.

What this lands:

- test/fakeagent/ is a new Go library:
  - fakeagent.go: Scenario, Event, Host, Assertion types with json tags. LoadScenario reads YAML via sigs.k8s.io/yaml (which goes through encoding/json so the same tags serve both authoring and wire emission). Validate enforces the structural invariants consumers rely on (name, host.id, non-empty timeline, known event_type). A Duration wrapper accepts "10ms"/"5s" strings or raw nanosecond integers.
  - feeder.go: Envelopes(opts...) materialises wire envelopes deterministically from the timeline (pure computation, no I/O); FeedControlPlane posts one envelope at a time to the headless binary's unix socket; PostDirect batches envelopes to a server's /api/events with a bearer token. Both honour the run config.
  - option.go: WithStartTime / WithHostID / WithSpeed / WithBatchSize / WithIDGenerator. Default speed is 0 (fire immediately), default batch size 100, default id generator is crypto/rand 32-hex.

- test/fakeagent/scenarios/ ships three starter YAML files: quiet-host.yaml (heartbeat-only noise floor), exec-fork-exit.yaml (baseline process tree), dns-and-network.yaml (DNS + outbound TCP, RFC 5737 documentation IP so the scenario reads as synthetic).

- Unit tests cover the loader (round-trip across the starter corpus, every validation branch, file-not-found, malformed YAML, Duration parsing both forms with invalid inputs), envelope generation (deterministic timestamps, host id override, payload-shape per event type cross-checked against schema/events.json, timeline-order preservation), and both feeders (stub unix-socket control plane records bodies, stub /api/events server records batches, error propagation, ctx-cancel mid-flight).

- shortSocketPath helper bypasses t.TempDir for unix sockets because macOS's sun_path is 104 bytes and long test names blow that budget; documented inline with a nolint:usetesting note.

[...]

PR #211, opened 2026-05-20
M2: headless agent binary with unix-socket control plane

getvictor:
Second milestone of the UAT plan in docs/testing-strategy.md. With M1's build-tag split letting the agent's Go core compile on linux, M2 adds a dedicated binary that wires the agent's queue + uploader + enrollment pipeline and exposes a local unix-socket control plane so test scenarios can inject events directly into the receiver. This is the substrate the L3 headless-agent + server CI job (M3) will drive.

What this lands:

- agent/receiver/receiver_other.go gains Inject([]byte) error on the non-darwin / no-cgo stub Receiver. Non-blocking: a full buffer returns the new ErrBufferFull sentinel so a hung uploader surfaces as a test failure rather than a hung scenario driver. The darwin XPC receiver has no Inject method by design; production XPC events come from the kernel via the system extension.

- agent/cmd/fleet-edr-agent-headless/ is the new binary, all four source files tagged //go:build !darwin || !cgo so production darwin builds skip the whole package:
  - main.go: thin entrypoint, parses --control-socket flag, loads config, runs the production enrollment.Ensure flow, hands off to Run. Excluded from Sonar coverage via the existing **/cmd/**/main.go rule.
  - headless.go: Options + Run. Mirrors the production agent's wiring minus the XPC-specific goroutines (no process reconciler, no commander, no XPC heartbeat). Bounded drain on shutdown so a hung server cannot block exit.
  - control.go: unix-socket HTTP server with POST /event (inject one JSON envelope) and GET /state (events_injected, inject_errors, last_inject_at_unix, queue_depth). 1 MiB body cap, 0600 socket perms, 5s read/write timeouts.
  - headless_test.go: integration test that stands up an httptest fake EDR server, runs Run in a goroutine, posts events via the unix socket, and asserts they arrive at the fake server with the right bearer token. Plus focused unit tests for Options.validate and the handler's error branches (empty body, invalid JSON, body too large, buffer full). Per-file coverage on new code is 71-100% (Sonar's 80% new-code gate has headroom).

- Taskfile.yml: build:agent:headless task that produces the linux binary under agent/tmp/. test:go:agent and test:go:agent:coverage now run a second pass under CGO_ENABLED=0 to cover the headless package whose build tags exclude it from the default cgo run; the two coverage profiles cover disjoint files (build-tag-enforced) so they merge cleanly by concatenation.

[...]

PR #214, opened 2026-05-21
M5: Playwright agent fixture + L4 smoke spec

getvictor:
Fifth milestone of the UAT plan in docs/testing-strategy.md. M1-M4 shipped the Go side of the agent + L3 integration test; M5 brings the same scenario corpus into the L4 browser-E2E layer so UI tests can seed realistic host / event / process state without direct-SQL hacks like fixtures/db.ts's seedCriticalAlert.

What this lands:

- test/e2e/fixtures/agent.ts is a new Playwright fixture. It reads the M3 YAML scenarios from test/fakeagent/scenarios/ (single source of truth shared with the Go library), enrols a fresh host via /api/enroll with the dev server's dev-enroll-secret, materialises the timeline into wire envelopes matching schema/events.json byte-for-byte, then POSTs them to /api/events with the minted bearer token. The Playwright `test` import is extended with an `agent.runScenario(name, opts?)` method that returns {hostId, eventsPosted, hostToken}.

  Why direct POST rather than going through the M2 headless binary's control plane: the queue + uploader path is already L3-covered by M4 (test/integration/agentserver). L4 cares about UI behavior under realistic data, not queue mechanics. Direct POST avoids process orchestration in playwright.config.ts and per-test host enrollment is much cleaner than the headless binary's single startup-time host_id.

  Why TypeScript rather than a Go/Node FFI bridge: the envelope shaping is small and stable (seven event types, ~80 LOC). The YAML scenarios are the contract; the mapping logic can sit in either language. Maintaining a shared mapping library is cheaper than maintaining the bridge.

- test/e2e/tests/qa/agent-events-flow.spec.ts is the L4 smoke spec. Four subtests: one per starter scenario (quiet-host / exec-fork-exit / dns-and-network) asserting the right event_types land in the events table for the per-test minted host_id, plus one that proves hostIdOverride wins over the scenario's host.id. Asserts via direct MySQL query (the same path fixtures/db.ts uses for the existing specs) rather than session-gated REST queries, to skip the operator-login ceremony.

  State isolation: each test mints a fresh UUID via crypto.randomUUID and passes it as hostIdOverride, so tests don't collide on host_id and the spec doesn't need a global DB wipe in beforeEach (the alert_events -> events FK constraint makes blanket DELETE awkward; unique host_ids keep the assertion accurate against accumulated rows from prior runs).

- test/e2e/package.json: adds js-yaml + @types/js-yaml (tiny, well-tested, what fixtures/agent.ts uses to parse the shared scenario corpus).

- docs/testing-strategy.md L4 section: now describes the fixture as it actually ships, including the rationale for not routing through the headless binary.

Verified locally: npx playwright test
tests/qa/agent-events-flow.spec.ts runs all 4 subtests against the existing dev server (started by Playwright's webServer block) in under 1 second. npx tsc --noEmit clean, task lint:md clean.

Existing specs (auth/break-glass, qa/reauth-modal-retry, etc.) are untouched - they still use fixtures/db.ts's direct-SQL helpers because seeding alert rows via /api/events requires rule-firing scenarios that don't exist in the corpus yet (M10 territory). The M5 fixture is for new specs and incremental migrations as rule- firing scenarios land.

Builds on M1 (#206), M2 (#211), M3 (#212), all merged. Independent of M4 (#213) which is still open. Refs UAT plan M5 milestone. M6 (UI specs expanded - host list pagination, process-tree fork bomb, alert detail per rule) is next.

## Summary by CodeRabbit

* **Documentation**
  * Updated E2E testing strategy documentation with detailed guidance on agent fixture testing.

* **Tests**
  * Added agent event simulation fixture for E2E tests to seed event data from scenario files.
  * Added test scenarios validating agent event flow and persistence.
  * Added test dependencies for scenario file parsing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/214?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/testing-strategy.md
+++ b/docs/testing-strategy.md
@@ -89,7 +89,12 @@ A run at any layer implies all lower layers have already passed; CI gates enforc
 - Playwright tests live in `test/e2e/`. Real server + real MySQL + real UI; tests start the server via the
   `webServer` block in `playwright.config.ts`.
 - For tests that need realistic agent data (process trees, alerts, app-control flow, host list), the spec drives the
-  fake-agent control plane to inject a scenario, waits for the events to land, then drives the UI.
+  M5 fixture at `test/e2e/fixtures/agent.ts`: it reads the same YAML scenarios as the Go fakeagent library
+  (`test/fakeagent/scenarios/`), enrolls a per-test host via `/api/enroll`, generates wire envelopes matching
+  `schema/events.json`, then POSTs them to `/api/events` with the minted bearer token. The smoke spec at
+  `test/e2e/tests/qa/agent-events-flow.spec.ts` proves the wire end-to-end across the starter corpus. The fixture
+  bypasses the headless agent binary intentionally - the queue + uploader path is already L3-covered (M4); L4 is
+  about UI behavior under realistic data.
 - For pure auth/RBAC/session tests, direct SQL fixtures under `test/e2e/fixtures/db.ts` remain the right tool; they
   are faster and the data they seed is not exercised through the ingestion path.
 - CI runs Playwright in phases (`scripts/test-e2e-coverage.sh`) so env-isolation tests (rate limits, short session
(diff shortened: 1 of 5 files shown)


getvictor:
@coderabbitai full review

PR #264, opened 2026-05-24
M13 markers agent xpc receiver (#250)

### LINKED ISSUE #250: agent-xpc-receiver: no Go test coverage; M13 markers blocked
## Coverage gap

`agent/receiver/` has Go source files but no `_test.go` file. The package is gated by `//go:build darwin && cgo`
(see `receiver.go:1`, `callbacks.go:1`), and its production path goes through a CGo bridge in
`agent/xpcbridge/xpc_bridge.{c,h}` to talk to the macOS XPC services exposed by the system extension and
the network extension.

The `agent-xpc-receiver` spec defines 10 scenarios spanning connection lifecycle, exponential backoff,
parallel receiver loops, disconnect tolerance, and clean shutdown
(`openspec/specs/agent-xpc-receiver/spec.md`). None can be marked under M13 today because there is no test
to attach a `// spec:<id>` comment to.

## Why not roll it into a marker bundle

Writing the missing tests is not a documentation pass:

1. **CGo build constraint**: the production receiver only compiles on `darwin && cgo`. CI Linux runners
   skip the package entirely. Tests would need either build tags + a macOS runner, or an injectable
   interface that lets the test bypass CGo.
2. **XPC mocking**: the bridge calls real `xpc_connection_*` Foundation APIs. Stubbing requires either a
   protocol abstraction (introduce `type Bridge interface { Connect(...) ... }` and pass a stub) or a
   C-side mock that returns canned responses.
3. **Concurrency**: every scenario involves connection state machines, backoff timers, reconnect loops.
   Deterministic tests will need clock injection + channel-based wait helpers similar to what
   `agent/reconcile/` already does.

Realistic estimate: 1-2 days of work to introduce the interface, port `Receiver` to use it, and ship 10
tests. Out of scope for any marker-rollout PR.

## What to do

Treat this as a separate work item from the M13 marker bundles. Once the test surface exists, marking
the 10 scenarios is a 30-minute commit. The marker work for M13 should NOT block on this; we'll close
the M13 `--strict` gate with the 10 scenarios still showing as uncovered until the test harness lands.

getvictor:
Extract reconnect/backoff/heartbeat from cmd/fleet-edr-agent into a testable Loop in agent/receiver. Production *Receiver now satisfies a Connector interface; the Loop drives it with a ConnectorFactory and exposes Sleep / Now hooks so backoff progression is observable in tests without real time.

Adds receiver.Dispatcher (atomic.Pointer over a connector box) that replaces the prior main.go appControlDispatcher, wired via LoopHooks.OnConnected and OnDisconnected so the commander's application_control push routes to the live connection and fails fast with ErrNoConnector during reconnect windows.

Test surface (9 of 10 agent-xpc-receiver scenarios now marked):
- two-parallel-receiver-loops (both branches)
- exponential-reconnect-backoff (both branches; verifies 1s -> 2s -> 4s -> 8s -> 16s -> 30s cap, and reset-on-success)
- auto-reconnect-after-extension-restart
- dropped-events-during-disconnect-are-tolerated
- outbound-policy-push-routed-to-active-connection (both branches)
- clean-shutdown-on-context-cancellation

The remaining downstream-consumer-falls-behind scenario lives at the Receiver layer's CGo onEvent callback and needs an event-drop helper extraction to be testable without a live XPC peer; deferred as a small follow-up.

Heartbeat tests moved with the heartbeat goroutine from agent/cmd/fleet-edr-agent/heartbeat_test.go to
agent/receiver/heartbeat_test.go and continue to cover the five ctx-cancel / done-close / ping-success / ping-failure / failed-chan-full cases against the Loop's unexported runHeartbeat method.

QA on edr-dev VM (192.168.64.5):
- Old agent shut down cleanly, drained 145-event queue
- New agent boots, persists token, commander + reconciler start
- Both ESF + network Loops attempt their XPC services with observable backoff approaching the 30s cap (pre-existing sysext-side issue exercises the failure path naturally)
- SigNoz traces confirm POST /api/events from the VM returns 200 on the redeployed agent

Closes #250.

## Summary by CodeRabbit

* **Refactor**
  * Restructured receiver lifecycle management with new abstraction layer for connection handling, heartbeat monitoring, and reconnection logic.
  * Improved connection state management through a dispatcher system for routing outbound communications.

* **Tests**
  * Added comprehensive test suites for heartbeat behavior, connection lifecycle, exponential backoff logic, and dispatcher functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/264?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


getvictor:
@coderabbitai full review

PR #279, opened 2026-05-26
Address SonarCloud global findings (21 issues triaged)

getvictor:
17 fixed via code edits, 4 accepted via Sonar status change (intentional + matches existing convention or opt-in/test-only by design).

Code edits

agent/cmd/fleet-edr-agent-headless/headless/control.go (S1192 ×2): extracted "Content-Type" + "application/json" to contentTypeHeader + contentTypeJSON constants. Three call sites (POST /event ack, GET /state response, error envelope) all use the constants.

agent/cmd/fleet-edr-agent-headless/headless/headless.go (S1186): the `shutdownControlPlane := func() {}` default is now documented with an inline body comment explaining why the no-op exists (replaced by startControlPlane's real shutdown closure when SocketPath is non-empty).

agent/queue/queue_test.go (S3776): TestQueue_CapEvictionRacesInflightUpload had cognitive complexity 16 vs 15 allowed. Factored out capRaceTestPayload + capRaceEnqueueInflight + capRaceAssertInflightEvicted helpers. The IN-clause in the eviction probe is now composed dynamically from len(inflightIDs) so a future count change doesn't require a parallel SQL edit (the previous hardcoded 5-placeholder shape was a footgun the require.Len lint check papered over).

server/detection/internal/intake/handler.go (S3776): ParseAndValidateIngestBody complexity 17 vs 15. Extracted readArrayOpen + streamDecodeEvents + readArrayCloseAndEOF helpers; the parent function is now a 3-step orchestrator.

server/detection/internal/intake/handler_fuzz_test.go (S3776): assertValidOutput complexity 23 vs 15. Split the 200- branch into assertOKContract and the 400-branch into assertBadRequestErrCode; the parent switch is now linear.

test/scale/runner_headless.go (S1186): scaleTokenProvider.OnUnauthorized empty body now carries an inline comment inside the braces (the rule rejects a preceding doc comment alone).

scripts/uat/lib/common.sh (S7682 ×3 + S7679 ×1): added explicit `return 0` at the end of uat_log + uat_ssh_args; added `local tag` for the positional parameter in uat_fail rather than reading $1 directly. Bash syntax-checked clean.

extension/edr/extension/ApplicationControlStore.swift (S1075): extracted the hardcoded "/var/db/com.fleetdm.edr/application-control.json" URI into a `defaultStoragePath` static constant. The init's default argument now references the constant so the URI is named in one place + future Sonar S1075 hits on this surface land on the constant rather than the function signature.

test/e2e/fixtures/auth.ts (S7763): switched `export { uninstallVirtualAuthenticator }` to `export { uninstallVirtualAuthenticator } from "./webauthn"` (`export...from` style; dropped the dangling import).

test/e2e/tests/qa/host-list-and-process-tree.spec.ts (S2871 ×2) + test/e2e/tests/qa/agent-events-flow.spec.ts (S2871 ×2): replaced bare `.sort()` with `.sort((a, b) => a.localeCompare(b))` so the alphabetical comparison is locale-stable rather than relying on JavaScript's default UTF-16 code-unit ordering (which doesn't behave alphabetically for accented chars).

Accepted via Sonar status change (intentional, documented)

agent/metrics/metrics.go:31 godre:S8196 (QueueDepthSource interface naming) - matches the project's `<Subject>Source` convention for read-only callback contracts (precedent: server/metrics's GaugeSource). Renaming to `DepthReader` (the rule's preferred shape) would create a one-off inconsistency reviewers have to re-resolve every time they touch the metrics surface.

test/scale/runner.go:689 go:S4830 + go:S5527 (TLS server-cert + hostname verification disabled) - the `InsecureSkipVerify: true` branch is the documented opt-in for `task uat:scale --insecure-tls=true` against dev:server's mkcert cert. The default is `false`; production scale runs against a properly-certificated server never take this branch. The `//nolint:gosec` comment already documents this; Sonar's NOSONAR pragma wasn't recognised so explicit status change is the path that sticks.

test/e2e/fixtures/auth.ts:14 typescript:S2068 (hardcoded password) - BG_PASSWORD is a literal in a Playwright fixture, shared across L4 specs so test failures don't have to chase the password value across files. It's never used in production code; the file is build-tagged out of the production bundle. Pinning the literal to an env var would only relocate the data without changing its sensitivity.

Gates: go test ./agent/... ./server/... green; task lint:go clean; spectrace 266/266 strict; bash -n scripts/uat/lib/common.sh clean. Sonar global open-issue count drops from 21 → 0 after this PR's analysis lands (17 closed by code fix + 4 closed by accept).

### DIFF
--- a/agent/cmd/fleet-edr-agent-headless/headless/control.go
+++ b/agent/cmd/fleet-edr-agent-headless/headless/control.go
@@ -37,6 +37,11 @@ const (
 	// controlSocketMode locks the unix socket to the running user only. The headless binary is single-user dev/test tooling; a more
 	// permissive mode would let any process on the host inject events.
 	controlSocketMode os.FileMode = 0o600
+
+	// HTTP header names + content types centralised so Sonar's S1192 (duplicated literal) doesn't trip across the three
+	// inject + state + error response paths. None of these are wire contracts that vary per call site.
+	contentTypeHeader = "Content-Type"
+	contentTypeJSON   = "application/json"
 )
 
 // startControlPlane binds a unix socket at socketPath and serves the control-plane HTTP API on it. Returns a shutdown function the
@@ -138,7 +143,7 @@ func handlePostEvent(
 	n := cnt.eventsInjected.Add(1)
 	cnt.lastInjectAtUnix.Store(time.Now().Unix())
 	logger.DebugContext(r.Context(), "control plane injected event", "events_injected", n)
-	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set(contentTypeHeader, contentTypeJSON)
 	w.WriteHeader(http.StatusAccepted)
 	_ = json.NewEncoder(w).Encode(map[string]any{"events_injected": n})
 }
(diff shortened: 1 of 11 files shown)


getvictor:
@coderabbitai full review

Issue #208, opened 2026-05-20
App control: BINARY rule type silently allows the first exec of any binary (cold-cache bypass)


getvictor:
## Summary

`extension/edr/extension/ESFSubscriber.swift:108-154` (handleAuthExec) returns `ES_AUTH_RESULT_ALLOW` whenever the
`FileHashCache` does not yet have a SHA-256 entry for the exec target. The fill is kicked off the AUTH callback
(`FileHashCache.shared.startLazyFill(...)` at line 151) and the second exec of the same `(dev, inode, mtime)` tuple
catches the rule. The first exec does not.

This is documented in `ESFSubscriber.swift:54-59, 86-107` as "first launch allowed, second launch blocked", inherited
from the demo-cut posture. With v0.1.0 going out as a real release (not a demo), the cold-cache ALLOW is no longer
defensible as the steady-state posture for Application Control.

## Why this is a real bypass primitive

Application Control's purpose is to deny an exec the operator marked as bad before it runs. "Allow the first exec and
catch the second" defeats that for any attacker who only needs to win one execution, which is the dominant case for
malware: dropper runs once, persists, reboots into the persistence path. By the time the cache is filled, the
process tree the rule was supposed to prevent is already established.

The bypass is also trivially deterministic: an attacker who knows the EDR is deployed can use `(dev, inode, mtime)`
mutation (rename the binary, rewrite mtime, atomic-replace) to invalidate the cache key on every exec and keep
returning to the cold-cache ALLOW path indefinitely.

## What Santa does

Santa hashes synchronously inside the ES auth callback, guarded by a deadline (`SNTEndpointSecurityManager`'s exec
handler runs the SHA-1/SHA-256 compute on the callback thread, with a fallback to the configured
`FailClosed`/`FailOpen` posture if the deadline approaches). The result is that Santa's first exec is decided on the
real identity, not on an empty cache.

Santa's design choice is intentional: a deadline-bounded sync hash is the only honest way to enforce identity-based
rules on the kernel's first-look event.

## Proposed direction

Switch BINARY rule evaluation to a deadline-guarded synchronous hash:

1. Read the ES deadline from `es_message_t.deadline`.
2. Compute SHA-256 on the callback thread with a budget headroom (e.g. deadline minus 500ms safety margin).
3. If the hash completes in time, look up the snapshot and respond.
4. If it does not, apply a configurable fallback posture: fail-closed (DENY) for high-assurance pilots, fail-open
[...]

Issue #207, opened 2026-05-20
App control: CDHASH doc comment misattributes a hardened-runtime gate to Santa


getvictor:
## Summary

`server/rules/api/types.go:157-158` says:

```go
// RuleTypeCDHash matches against the signed Code Directory hash. 40 lowercase hex characters; only honored for hardened-runtime
// processes per the spec.
```

This presents a "hardened-runtime only" CDHASH rule as the spec, and elsewhere the same rationale gets attributed to
Santa. Santa does not gate CDHASH rules on hardened runtime. Santa's binary authorization engine fires CDHASH rules
against any code-signed binary (Apple platform binaries, third-party signed apps, ad-hoc signed binaries with a
CodeDirectory). Northpole's documented binary-authorization model lists CDHASH as a peer of BINARY / SIGNINGID /
CERTIFICATE / TEAMID, not as a hardened-runtime-only rule. Real-world Fleet usage already ships CDHASH rules against
non-hardened apps (e.g. the WhatsApp.app Santa profile in fleetdm/fleet PR #35242).

There may still be a defensible internal-policy reason to honor CDHASH only on hardened-runtime binaries here
(non-hardened binaries can be modified, weakening the integrity guarantee that makes CDHASH worth choosing over
BINARY). That is a fine choice. But the comment claims it is Santa's rationale and the spec's rationale, and neither
is true. The doc reads as if the author has not read Santa.

## Concrete risk

Quoting this rationale on stage, in a spec review, or in a customer conversation that includes anyone who has read
Santa's source (e.g. Patrick Wardle) makes us look like we are pattern-matching from a misread of the reference
implementation. It is a credibility hit on the Application Control claim.

## Proposed fix

Rewrite the comment to state the EDR's own rationale, separate from Santa's behavior. Two acceptable framings:

1. State the policy honestly. EDR honors CDHASH rules only on hardened-runtime processes because CDHASH's integrity
   guarantee depends on the kernel rejecting modifications, which non-hardened binaries do not enforce. This is a
   stricter posture than Santa, which fires CDHASH rules on any code-signed binary.
2. Drop the gate to match Santa. Remove the hardened-runtime constraint entirely, fire CDHASH on any code-signed exec
   where the kernel populated `target.cdhash`, and update the comment to say so.

(2) is the closer-to-Santa choice and is preferred if there is no concrete attack model that makes (1) worth the
incompatibility. But the immediate fix here is the comment, not the code.

## Scope

[...]

PR #289, opened 2026-05-26
App Control: platform-binary carve-out, deadline-guarded BINARY hash,…

### LINKED ISSUE #205: App control: failsafe carve-out has no critical platform binary allowlist
## Summary

The AUTH_EXEC failsafe in `extension/edr/extension/ESFSubscriber.swift:117-121` checks only `teamID == extensionTeamID`
(`"FDG8Q7N4CC"`). Anything not signed by our Apple team ID falls through to the snapshot lookup and is fully blockable
by a BINARY rule. That includes Apple platform binaries like `launchd`, `xpcproxy`, `fseventsd`, `kextd`, `sysextd`,
`systemextensionsd`, `WindowServer`, `loginwindow`, `mds`, and so on.

An admin who adds the SHA-256 of any of these to the rules table (deliberately, accidentally, or via a compromised
upstream feed) can render a Mac unbootable. The agent / extension / host app are protected because of the team_id
match, but the OS itself is not. This is a customer-bricking blast-radius bug that v0.1.0 ships into ten-to-five-hundred
endpoint pilots.

Reviewer's comparison point: Santa keeps a hardcoded critical-platform-binary allowlist (`SNTRuleTable` consults
`PrioritizedRuleType` + Apple-signed platform binaries before BINARY rules can fire). They do this exactly because
"admin adds wrong hash, customer's fleet bricks itself" is the unforgiving failure mode of binary authorization.

## Repro sketch

1. Identify the SHA-256 of `/sbin/launchd` on the target macOS.
2. POST it as a BINARY block rule to the policy via the REST surface.
3. Wait for snapshot sync to the host.
4. Next launchd exec (any new login, reboot, sysextd restart) hits the block.

The team_id failsafe does NOT save you here. Apple platform binaries are signed with Apple's identity, not ours.

## Proposed minimum bar for v0.1.0

Add a second carve-out clause in `handleAuthExec` before the snapshot lookup:

```swift
if target.is_platform_binary {
    es_respond_auth_result(client, message, ES_AUTH_RESULT_ALLOW, true)
    return
}
```

`es_process_t.is_platform_binary` is the kernel's own "this binary is on the Apple-signed system image" flag, which
already correctly identifies launchd/xpcproxy/fseventsd/kextd/sysextd/etc. It is more conservative than maintaining a
hand-curated path or signing-id allowlist and is the floor Santa effectively encodes.

(Note: passing `cache: true` on this ALLOW is intentional. Same identity, same answer, every time. See the separate
issue on the kernel result cache.)

## Stretch goal (can stay on v0.2.0)

A server-pushed allowlist on top of the platform-binary floor, so operators can extend it (Sophos kexts, MDM agents,
backup utilities) without an agent re-release. This is the Phase B failsafe list already alluded to in
`ESFSubscriber.swift:11-12, 91-94`.

## References

- `extension/edr/extension/ESFSubscriber.swift:7-13` (the extensionTeamID constant + comment)
- `extension/edr/extension/ESFSubscriber.swift:86-154` (handleAuthExec, where the failsafe runs)
- #187 (closed) shows ESF redacts team_id for ad-hoc-signed binaries, which means even our own failsafe doesn't fire in
  the dev VM. The platform-binary check is independent of that bug.

## Milestone rationale

v0.1.0. Customer-bricking failure mode. The fix is one line plus a test against `target.is_platform_binary` paths.
Cannot ship pilots with this gap.

getvictor:
… fallback posture

Closes #205, #207, #208.

#205 customer-bricking failsafe gap: handleAuthExec ALLOWs any exec the kernel classifies as target.is_platform_binary BEFORE the snapshot lookup, with cache:true so the kernel pins the result. An admin who pastes the SHA-256 of /sbin/launchd, xpcproxy, fseventsd, etc. into a BINARY block rule can no longer brick the host.

#208 BINARY rule cold-cache first-exec bypass: the lazy-fill ALLOW path is gone. handleAuthExec computes SHA-256 synchronously on the AUTH callback thread under a budget bounded by es_message_t.deadline minus a 500 ms safety margin. The decision walk is extracted into a pure-logic AuthExecDecider so the SwiftPM test target (which excludes ESFSubscriber. swift because of the EndpointSecurity import) can drive the verdict matrix. When the hash cannot complete, the snapshot's deadline_fallback posture decides:
  fail-closed (default): DENY + emit application_control_undecided.
  fail-open:             ALLOW silently (demo-equivalent).
  audit-only:            ALLOW + emit application_control_undecided.
The posture field rides on SetApplicationControlPayload's wire shape
(json key deadline_fallback). v0.1.0 has no DB column yet; every snapshot
ships with the default fail-closed. Per-policy persistence + REST surface
to set the value land in a v0.1.x follow-up.

#207 Santa attribution on CDHASH doc comments: Santa does NOT gate CDHASH on Hardened Runtime; we do, for our own reasons (page mapping is lazy and not re-verified post-load, so ESF's reported CDHash is not a reliable identity for the bytes that will eventually execute). Updated both comments to own the rationale and call out the Santa divergence so a migrating Santa admin is not surprised when a CDHASH rule no-ops against a non-hardened target.

Tests
  AuthExecDeciderTests.swift: 15 cases across the (3 postures x 3 hash
    outcomes + computed/notNeeded paths x 4 precedence layers) matrix.
  FileHashCacheTests.swift: 5 cases for lookupOrComputeWithDeadline
    (computed, warm cache hit, deadline exceeded, missing file, TOCTOU
    inode mismatch).
  app_control_wire_test.go: deadline_fallback default + per-posture
    passthrough + JSON key pin + IsValidFallbackPosture validator.
  All 104 swift tests + the touched Go suites + golangci-lint + xcodebuild
  for the extension target are green.

Behavior change for upgrades
The cold-cache first-exec ALLOW is removed. Pilots upgrading from a v0.1.0-rc.* see either a DENY (the fail-closed default) or a small latency hit (tens of ms on typical multi-MB binaries; bounded by the 500 ms safety margin) on the first exec of any unhashed binary. The v0.1.x follow-up that wires audit-only into the REST surface lets operators measure the rate per fleet before flipping to fail-closed.

## Release Notes

* **New Features**
  * Added operator-configurable deadline-fallback behavior (fail-closed, fail-open, audit-only) for application control rule evaluation.
  * Implemented rule precedence evaluation with automatic platform binary carve-out.
  * Added deadline-aware hash computation with timeout handling.
  * Introduced undecided audit events for cases where rule verdicts cannot be determined.

* **Documentation**
  * Updated operations guide with detailed application control policy model and rule evaluation runbook.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/289?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,6 +3,7 @@ included:
 
 excluded:
   - extension/edr/build
+  - extension/edr/.build  # SwiftPM-generated runner.swift carries vendored SwiftLint violations we cannot edit.
 
 opt_in_rules:
   - closure_body_length
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -391,12 +391,68 @@ be used.
 
 ## Application control
 
-Application Control (the replacement for the singleton blocklist) is
-introduced in the `add-application-control` OpenSpec change. Phase 1
-deletes the legacy `/api/policy` endpoints and the `set_blocklist`
-command; the new REST surface, agent command, and decision engine land
-in later phases of that change. Until then there is no in-product
-allow/block surface.
+Application Control replaces the singleton blocklist with operator-managed
+policies (named, versioned, audited) containing typed rules. The AUTH_EXEC
+handler in the system extension consults the active snapshot on every exec
+and denies the first matching rule with `action=BLOCK` and
+`enforcement=PROTECT`. Precedence is CDHASH → BINARY → SIGNINGID → TEAMID.
+CERTIFICATE and PATH rules are accepted by the REST surface but deferred
+to Phase B (leaf-cert cache / Launch Services indirection).
+
+Two unconditional carve-outs run before any rule:
+
+1. **Platform-binary carve-out.** If the kernel sets
+   `target.is_platform_binary` (launchd, xpcproxy, fseventsd, kextd,
+   sysextd, WindowServer, etc.) the handler returns ALLOW with the kernel
+   cache pinned. An admin who pastes the SHA-256 of `/sbin/launchd` into a
+   BINARY rule will NOT brick the host: the carve-out fires before the
+   snapshot walk.
+2. **Self-allow failsafe.** The agent / extensions / host app (matched by
+   both Fleet's team_id and the exhaustive Fleet bundle-id allowlist)
+   ALLOW unconditionally. A misconfigured rule cannot block the EDR
+   itself.
+
+### BINARY rules and the deadline-fallback posture
+
+A BINARY rule matches the file's SHA-256. The hash is computed
+synchronously on the AUTH_EXEC callback thread, with the budget bounded
+by the kernel-supplied `es_message_t.deadline` minus a 500 ms safety
+margin for the post-hash work. v0.1.0 closes the pre-RC bypass primitive
+where the first exec of any binary slipped past BINARY rules while the
+cache filled asynchronously (#208).
+
+If the hash cannot finish in time (large binary, slow disk, or the file
+is unreadable due to a TOCTOU replace between AUTH and read), the
+snapshot's `deadline_fallback` field drives the verdict:
+
+| Posture | Verdict on deadline / read failure | Event emitted | When to pick |
+|---|---|---|---|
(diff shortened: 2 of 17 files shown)


getvictor:
@coderabbitai full review

PR #291, opened 2026-05-27
Detection: operator per-rule disable via EDR_DISABLED_RULES (#238)

### LINKED ISSUE #238: Detection engine: implement operator per-rule disable (spec/impl gap surfaced by M13 markers)
## Background

Surfaced during M13 marker work for \`openspec/specs/server-detection-rules-engine/spec.md\`. The spec declares the following normative requirement:

> ### Requirement: Operator toggling of individual rules
> 
> The system SHALL allow an operator to disable individual rules at startup through configuration. A disabled rule MUST NOT
> evaluate against any batch and MUST NOT produce alerts until it is re-enabled.

No such mechanism exists today. The current operator-tuning surface in \`server/rules/api/RegistryOptions\` and \`server/rules/internal/catalog/registry.go\` carries only **allowlists** (\`SuspiciousExecParentAllowlist\`, \`LaunchAgentAllowlist\`, \`LaunchDaemonTeamIDAllowlist\`, \`SudoersWriterAllowlist\`) — operators carve out specific binaries/paths/team-IDs that won't trip an individual rule, but cannot disable a whole rule.

Closing this gap is the only blocker on covering scenario \`server-detection-rules-engine/operator-toggling-of-individual-rules/an-operator-disables-a-noisy-rule-for-their-environment\` under M13's full-gate mode (#230). Today the scenario is the lone uncovered SHALL in this spec; the surrounding ten are covered by the M13 first-batch PR.

## Acceptance

1. \`api.RegistryOptions\` grows a \`DisabledRuleIDs []string\` (or equivalent) field that catalog.New filters against when assembling its return list.
2. \`server/config/config.go\` plumbs an \`EDR_DISABLED_RULES\` env var (comma-separated rule IDs) through to RegistryOptions at boot.
3. \`cmd/fleet-edr-server/main.go\` passes the resolved RegistryOptions to \`rulesbootstrap.New\` so the engine's LoadActive only sees enabled rules.
4. Test in \`server/detection/internal/tests/integration_test.go\` exercising the spec scenario: configure the registry with a stub rule disabled, post a batch that would otherwise trigger it, assert no alert. The test marker is the canonical ID above.
5. \`docs/operator-runbook.md\` (or equivalent) documents the \`EDR_DISABLED_RULES\` knob with an example.

## Design notes

- A disabled rule should not show up in \`Engine.Catalog()\` either — otherwise \`tools/gen-rule-docs\` will document a rule that doesn't fire, which is operator-misleading. The right boundary is at \`catalog.New\`: filter the slice before returning.
- The disable list is **boot-time only** per the spec. Hot-reload of rule enable/disable can be a separate change.
- Unknown rule IDs in the disable list should log a warning (not fail boot) so a stale operator config doesn't take down the server.

## Related

- M13 first context PR: #235
- M13 marker backlog umbrella: #230
- Existing allowlist tuning surface: \`server/rules/api/types.go\` \`RegistryOptions\`
- Surfaced this gap: this PR (M13 server-detection-rules-engine markers)

getvictor:
Closes #238. Closes the last uncovered SHALL in the server-detection-rules-engine spec by wiring a boot-time disable knob through the rule catalog.

api/RegistryOptions
  Adds DisabledRuleIDs []string. catalog.New drops any rule whose
  ID() appears in this slice, so a disabled rule is gone from the
  engine's active set AND from Engine.Catalog() (the GET /api/rules
  surface and tools/gen-rule-docs both stop listing it).

catalog
  - New(opts) now filters opts.DisabledRuleIDs after constructing the canonical 9-rule slice. Zero-value opts keeps the existing "all rules registered" behaviour.
  - UnknownDisabledIDs(opts) is a small diagnostic helper that returns the entries of DisabledRuleIDs that don't match any production rule. bootstrap.New uses it to warn at boot when a stale config references a renamed / removed rule; the boot succeeds either way so a typo doesn't take a deployment down (per #238 design notes).

config
  Adds EDR_DISABLED_RULES env var. Comma-separated rule_id values,
  whitespace trimmed, empty entries dropped. Same parse shape as the
  existing allowlist knobs.

cmd/fleet-edr-server/main.go
  Threads cfg.DisabledRuleIDs into the RegistryOptions struct literal
  alongside the existing allowlists.

Tests
  - registry_test.go: catalog-level unit tests cover the filter (single / multiple / all-disabled / unknown-id) AND the UnknownDisabledIDs diagnostic helper.
  - server/rules/internal/tests/integration_test.go: end-to-end through rulesbootstrap.New proves Catalog().List() AND ContentService().ActiveRules() both honour DisabledRuleIDs. Spec marker: server-detection-rules-engine/operator-toggling-of-individual-rules/an-operator-disables-a-noisy-rule-for-their-environment
  - server/detection/internal/tests already has the engine-layer counterpart (TestEngine_OperatorDisablesNoisyRule) which covers the LoadActive-swap half of the same spec scenario.

Docs
  - install-server.md: env var table row.
  - operations.md: longer-form rationale comparing the disable knob to the existing allowlist knobs.

## Release Notes

* **New Features**
  * Introduced `EDR_DISABLED_RULES` configuration option to disable specific detection rules at server startup. Disabled rules are removed from the API, dashboards, and alert evaluation. Server restart is required for changes. Unknown rule IDs trigger a boot warning but do not prevent startup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/291?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


getvictor:
@coderabbitai full review

PR #304, opened 2026-05-29
Selective ESF persistence

getvictor:
## Release Notes

* **New Features**
  * Added detection of LaunchDaemon persistence via Background Task Management event registration; now triggered when unsigned executables are registered as system daemons rather than on plist file writes.
  * Added support for process-optional alerts, enabling detections that aren't tied to a specific process.
  * New application control block alert detection when extension denies executable execution based on admin-defined rules.

* **Bug Fixes & Improvements**
  * Alert deduplication now uses executable code-signing evaluation rather than instigator signing for LaunchDaemon persistence detection.
  * Refined alert severity for LaunchDaemon persistence from critical to high.
  * Improved handling of MDM-managed and platform-signed executables to reduce false positives.

* **Documentation**
  * Updated detection rules documentation and scenarios to reflect BTM-based LaunchDaemon persistence detection.
  * Added specification updates for process-optional alert behavior and subject-based deduplication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getvictor/fleet-edr/pull/304?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #301, opened 2026-05-29
Selective ESF subscription (ADR-0008): drop broad NOTIFY_OPEN, BTM for persistence, target-muted CREATE/WRITE


getvictor:
**Re-scoped per [ADR-0008](../blob/main/docs/adr/0008-selective-esf-subscription.md) (Selective Endpoint Security
subscription).** The original framing ("add `es_mute_path` to mute noise") doesn't survive contact with two ESF
constraints and an industry calibration:

- `NOTIFY_OPEN` **cannot be muted per-event-type** (`es_mute_path_events` silently ignores it), and
- target-path muting + inversion is **global across event types** - `AUTH_EXEC`'s target is the executable, so naively
  "selecting only `/Library/LaunchDaemons`" would mute exec auth + Application Control.

Calibration against Apple guidance + commercial-agent analysis + Red Canary Mac Monitor: top EDRs do selective
source-side subscription, avoid broad `NOTIFY_OPEN`, and detect launch-item persistence via `BTM_LAUNCH_ITEM_ADD`. See
ADR-0008 for the full reasoning and the two constraints.

## Goal

Move the extension from "subscribe to broad `NOTIFY_OPEN`, forward everything, server filters" to selective source-side
subscription, cutting the open/create firehose that caused the ~12-min detection backlog (finding-a / #300) and feeds
the #298 exec-stall.

## Plan (staged; each step VM-validated before the next)

1. **Persistence → BTM.** Detect LaunchDaemon/LaunchAgent/login-item persistence via
   `ES_EVENT_TYPE_NOTIFY_BTM_LAUNCH_ITEM_ADD`; migrate `privilege_launchd_plist_write` off the file-write proxy. Biggest,
   cleanest win; also closes the atomic-rename / platform-binary-`cp` gaps the file-write rule documents. (Decide:
   whether to retain a narrow target-muted CREATE/WRITE watch on `/Library/LaunchDaemons` for early-drop defense in
   depth, since BTM fires at registration, not at the drop.)
2. **`sudoers_tamper` → target-muted CREATE/WRITE.** Subscribe to `NOTIFY_CREATE`/`NOTIFY_WRITE` (+`RENAME`) with
   per-event-type inverted TARGET-path muting scoped to `/etc/sudoers` + `/etc/sudoers.d` only (never global inversion;
   call `es_unmute_all_target_paths` before inverting).
[...]

Issue #298, opened 2026-05-29
ESF AUTH_EXEC: synchronous hashing on the serial ES thread can stall (and fail-closed DENY) all execs under load


getvictor:
## Summary

During an in-place upgrade of the Endpoint Security system extension on a SIP-enabled macOS host (L5 system-test,
`edr-qa` VM, `v0.1.1-rc.2` / commit `69d83d0`), **new process creation stalled system-wide** for several minutes:
new SSH logins timed out at "banner exchange" (the point where sshd forks child processes), while the kernel (ping)
and already-established connections stayed fully responsive. The agent logged a storm of `receiver event channel full`.
A clean reboot (single extension running) cleared it and the stall did not recur.

## Root cause

The intuitive hypothesis (a full agent channel blocks the ES auth response) is **refuted** by the code, which is good:
the AUTH allow/deny verdict is decoupled from agent delivery, and a full agent channel cannot block `es_respond`. The
real mechanism is:

- The extension subscribes to `ES_EVENT_TYPE_AUTH_EXEC` (its only AUTH event) with a single `es_new_client` handler
  block; ES delivers AUTH messages **serially**.
- `handleAuthExec` performs **synchronous SHA-256 hashing** (`FileHashCache.computeSHA256WithDeadline`) inline on the ES
  delivery thread whenever the active ruleset contains BINARY rules. It is deadline-bounded but runs on the serial path.
- During an in-place upgrade the new extension's hash cache is **cold**, two ES clients briefly **coexist** (old
  terminating + new activating), and they contend for disk I/O. Each exec's AUTH decision then serializes behind a slow
  hash, stalling every exec on the machine.
- `AuthExecDecider.deadlineFallback` defaults to **failClosed** -> execs whose hash exceeds the deadline are **DENIED**,
  not merely delayed.

So the `channel full` storm and the exec stall share a common cause (the upgrade-time exec/event burst); the channel
drops are a co-symptom, not the cause.

## Evidence

- Agent log: repeated `receiver event channel full` (security extension XPC service) throughout the window.
- New SSH logins timed out at banner exchange for ~5+ minutes; `ping` and existing connections were unaffected.
- No recurrence after a clean reboot with only the new extension active.

## Code references

- `extension/edr/extension/ESFSubscriber.swift` - single `es_new_client` handler; `AUTH_EXEC` subscription; `handleAuthExec`.
[...]

PR #307, opened 2026-05-31
Selective ESF subscription: dedicated sudoers file-tamper client, drop broad NOTIFY_OPEN (#301)

getvictor:
Implements ADR-0008 steps 2-3 (part of #301; step 1 shipped in #304, step 4 deferred).

## What changes
- **`FileTamperSubscriber`** — a second, NOTIFY-only ES client (`NOTIFY_CREATE` + `NOTIFY_WRITE`). It calls `es_unmute_all_target_paths`, mutes `/etc/sudoers` (TARGET_LITERAL) + `/etc/sudoers.d` (TARGET_PREFIX) (and the `/private/etc` spellings ESF reports), then `es_invert_muting(ES_MUTE_INVERSION_TYPE_TARGET_PATH)` so it observes **only** those paths. Each event is re-emitted as a write-mode `open` event, so the server `sudoers_tamper` rule is unchanged. The inversion is client-global, so it lives on this auth-less client alone — the primary client's `AUTH_EXEC` / Application Control are untouched (the landmine ADR-0008 identified).
- **Primary client drops broad `NOTIFY_OPEN` + `NOTIFY_CREATE`** (the firehose that drove the ~12-min detection latency and fed #298). `sudoers_tamper` was the only `open` consumer; LaunchDaemon persistence moved to BTM in #304.
- **`NOTIFY_RENAME` deliberately not subscribed** — visudo/sudoedit write via temp-file + atomic rename onto `/etc/sudoers`, so watching rename would fire on every legitimate edit. The atomic-replace gap stays documented (unchanged). Step 4 (server-pushed watched-path set) remains future work.
- `sudoers_tamper.go` doc-only (collection mechanism + rationale); match logic unchanged. `endpoint-event-collection` spec + ADR-0008 updated. `ESFSubscriber+FileEvents.swift` removed.

## Validation
xcodebuild SUCCEEDED · swiftlint 0 · Go catalog tests · spectrace 269/269 · golangci-lint 0 · markdownlint 0.

**edr-dev (SIP-off, fully exercises the muting/inversion):** both clients start (primary "Subscribed to 5 event types", file client "CREATE/WRITE only"); sudoers CREATE+WRITE fire; a non-sudoers write does NOT (inversion scoping correct); zero broad `open` events (firehose gone); exec still flows. SIP-on App-Control-still-enforces confirmation pending an rc on edr-qa.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated, optimized monitoring for critical sudoers files (`/etc/sudoers` and `/etc/sudoers.d/`) to improve detection accuracy and reduce false positives.

* **Bug Fixes**
  * Eliminated false positives from legitimate sudoers edits by excluding rename operations from monitoring.


getvictor:
@coderabbitai full review

PR #290, opened 2026-05-27
App Control: wire CERTIFICATE + PATH rule types end-to-end (#210)

getvictor:
Closes #210. Phase B close-out of the original umbrella #68: every wire-enum rule type is now wired through the validator AND the extension's AUTH_EXEC decision walk. Precedence (Santa's order):
  CDHASH → BINARY → CERTIFICATE → SIGNINGID → TEAMID → PATH

Server validator
  - ValidateRuleType: CERTIFICATE + PATH accepted (was: returned ErrAppControlUnsupportedRuleType). Only unknown tokens stay invalid.
  - ValidateIdentifier: drops the trailing unsupported-error returns in the CERTIFICATE + PATH branches. Format checks unchanged.
  - validate_test.go: collapsed the deferred-types case to a single accepted-types table; explicit unknown-token row stays.
  - app_control_test.go: rewrote the unsupported-types test against an unknown-token submission; positive companion now covers all six wired types.

Extension
  - SigningInfoFallback: extended to return SigningInfo(teamID, leafCertSHA256). One SecCodeCopySigningInformation call now covers both fields; cache key unchanged. leafCertSHA256 reads kSecCodeInfoCertificates[0] and SHA-256s the DER bytes (matches `codesign -d --extract-certificates` + `shasum -a 256`).
  - AuthExecDecider: AuthTuple gains leafCertSHA256 + canonicalPath. decideAuthExec walks the new 6-layer precedence. Extracted the lower-precedence walk + the hash-outcome reason helper to keep cyclomatic_complexity under the SwiftLint cap. Added a pure canonicalizePath helper that mirrors the server's Go rules verbatim (reject empty / relative / `..`; rewrite /tmp, /var, /etc → /private).
  - ESFSubscriber: buildAuthTuple resolves leafCertSHA256 (one SecCode call shared with the TeamID path) + canonicalizePath of the exec target path. Top-of-file rationale comments refreshed.
  - CDHashHex.swift (new): isHardenedRuntime + cdhashHexString helpers extracted from ESFSubscriber.swift so that file stays under file_length cap.

Tests
  - AuthExecDeciderTests.swift: makeRule / makeSnapshot / makeTuple helpers lifted to file scope so the class body fits under type_body_length. Split CERTIFICATE + PATH cases into a sibling AuthExecDeciderPhaseBTests class. CanonicalizePathTests pins the Swift implementation against the Go canonicalizer's table.
[...]

PR #314, opened 2026-06-01
UAT: add L5 app-control-block scenario (active-blocking coverage)

getvictor:
Closes the coverage gap that let App Control active blocking ship without L5 validation: when /api/policy became the per-policy /api/v1/app-control/* surface (#289 / #290), the old blocklist L5 scenario was deleted and its replacement was only "tracked separately" -- so enforcement lost system/VM coverage and nothing forced it back. This is that replacement, and the missing half of the #301 acceptance gate ("Application Control still works", VM-validated).

scripts/uat/scenarios/app-control-block/:
- Builds a non-platform block target on the VM with `go build` (a copied system binary is arm64e and AMFI-killed when run outside SIP; a Go binary is plain arm64 + non-platform, the same trick attack-runbook uses for its launchd drop).
- Confirms it runs at baseline, POSTs a BINARY BLOCK rule on its hash to the Default policy, then polls until the exec is DENIED (host-side enforcement).
- Asserts the application_control_block alert itself: the alert rule_id is dynamic (app_control:<ruleId>), which the driver's static rule_id match can't express, so the scenario carries no rules block and the driver passes it on attack.sh exit 0 (a mode it explicitly supports). Cleans up the rule + target.

Validated on edr-qa (SIP + Gatekeeper on) against v0.1.1-rc.8: baseline allowed -> rule posted -> exec DENIED -> app_control:<id> alert (source=application_control) -> PASS.

Also scrubs the now-resolved "tracked separately" notes in scripts/uat/README.md, scripts/qa/README.md, and docs/testing-strategy.md to point at this scenario.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Documentation**
  * Updated testing strategy documentation to clarify Application Control blocking scenario coverage and test automation procedures
  * Added new UAT scenario documentation for validating Application Control active blocking behavior with expected alert assertions
  * Refreshed QA and UAT procedures to reflect current test infrastructure for application control testing

### DIFF
--- a/docs/testing-strategy.md
+++ b/docs/testing-strategy.md
@@ -134,9 +134,10 @@ A run at any layer implies all lower layers have already passed; CI gates enforc
   within the per-rule SLA. UAT plan milestone **M9** delivered the driver plus one starter scenario,
   `attack-runbook`, which asserts six rule_ids fire from the dogfood runbook. The scenario is a thin wrapper
   around `scripts/qa/attack-runbook.sh`; the dogfood script stays the interactive demo, M9 adds the asserted
-  automation layer on top. App Control gets its own UAT scenario tracked separately against the per-policy
-  `/api/v1/app-control/*` surface. Driver supports `--dry-run` for orchestration smoke-tests without driving
-  real infrastructure. See `scripts/uat/README.md` for the schema - capture procedure.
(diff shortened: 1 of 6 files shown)


getvictor:
@coderabbitai full review

Issue #178, opened 2026-05-16
Agent XPC connection silently fails after extension restart


getvictor:
## Problem

Discovered during issue #173 / #176 QA on edr-dev (192.168.64.5). After the security extension is killed and auto-restarted by sysextd, the agent's XPC connection to the new extension instance produces no "Peer connected" event on the extension's listener — even though:

- `launchctl print system/FDG8Q7N4CC.com.fleetdm.edr.securityextension` shows the Mach endpoint as `active=1, managed=1, port=0xd6bfb` and the extension PID is alive.
- The agent's receiver.New + xpc_connection_create_mach_service + xpc_connection_activate + send-hello all succeed silently. Agent logs `receiver connected service=FDG8Q7N4CC.com.fleetdm.edr.securityextension.xpc`.
- The extension's XPCServer listener queue is alive and processing events — confirmed by the issue-#173 phantom-peer logs.

Yet the extension's `XPCServer.handleListenerEvent` never receives `XPC_TYPE_CONNECTION` for the real agent peer. Events sent by the extension pile up in the pending-send buffer (added by PR #TBD for issue #173) until the 10k cap is reached, then get dropped one by one with a WARN.

edr-qa (192.168.64.7) is unaffected — its agent ingest pipeline keeps working across extension events. The bug appears specific to edr-dev's current state and persists across clean restart sequences (bootout agent → SIGKILL extension → sysextd respawn → bootstrap agent).

## Repro on edr-dev (current state)

```
# stop agent
sudo launchctl bootout system /Library/LaunchDaemons/com.fleetdm.edr.agent.plist
# kill extension (sysextd auto-restarts it)
sudo kill -9 $(pgrep -f securityextension)
sleep 8
# wait for Mach service to bind, then start agent
sudo launchctl bootstrap system /Library/LaunchDaemons/com.fleetdm.edr.agent.plist
# observe: agent log says "receiver connected"; extension log shows phantom Peer connect + disconnect within ~10ms; no real peer connect afterwards.
```

The phantom peer is *not* the real agent — its lifecycle is ~10ms and it produces no inbound hello. The real agent's connection silently goes to a black hole.

## Hypotheses (untested)

1. macOS XPC caches the Mach service binding by name; when sysextd re-registers the same service, existing client connections may stick to a stale port that no longer routes. The agent doesn't detect this because xpc_connection_send_message is fire-and-forget and never returns a delivery failure for a "service moved" case.
[...]

PR #331, opened 2026-06-08
Fix network extension delivering no events (missing hello-ack handshake)

getvictor:
The network extension's XPCServer never delivered network_connect/dns_query events to the agent: the agent's receiver logged "xpc_bridge_connect failed for ...networkextension.xpc" every reconnect cycle, while the security extension worked fine through the same agent.

Root cause: the agent's xpc_bridge_connect (agent/xpcbridge/xpc_bridge.c, issue #178) sends a {"type":"hello"} and cancels the connection + returns failure if no {"type":"hello-ack"} arrives within 5s. The security extension's XPCServer answers hello with hello-ack; the network extension's XPCServer only handled XPC_TYPE_ERROR on its peer connection and never read inbound dictionaries, so it never saw the agent's hello and never acked. The agent's network-extension receiver therefore timed out forever and no flow/DNS events were consumed. The NE server had drifted from the ES server when the hello-ack handshake (issue #178) was added only to ES.

Fix: the NE XPCServer now dispatches inbound XPC_TYPE_DICTIONARY messages and replies to hello with hello-ack (networkXPCShouldAck), matching the security extension and satisfying the existing extension-xpc-server "Hello handshake and reply" spec requirement (which applies to both extensions). The decision is extracted into a Foundation-only NetworkXPCInbound.swift so it is unit-testable via the SwiftPM logic facade (auto-compiled into the networkextension Xcode target via its synchronized file group) and pinned by a regression test.

Verified: swift build + swift test (16 logic tests, incl. the new NE handshake regression), swiftlint, spectrace 380/380. End-to-end (events actually flow on a VM with the rebuilt extension) still to confirm.

Follow-up: the ES and NE XPCServers are near-duplicate; extracting a shared XPC server would prevent this kind of drift. Tracked separately.

## Checklist

      (flagged below).

## Release Notes

* **New Features**
  * Enhanced network extension initialization with improved handshake protocol and error handling for peer communication.

* **Tests**
  * Added test coverage for network extension message validation and handling.

### DIFF
--- a/extension/edr/Package.swift
+++ b/extension/edr/Package.swift
@@ -97,7 +97,12 @@ let package = Package(
                 "extension/EventSerializer.swift",
                 "extension/FileHashCache.swift",
                 "extension/SigningInfoFallback.swift",
-                "networkextension/DNSParser.swift"
+                "networkextension/DNSParser.swift",
+                // Pure hello-handshake decision for the network extension's XPCServer. Foundation-only, uniquely named
+                // (NetworkXPCMessageType / networkXPCShouldAck) so it does not collide with the security extension's
+                // XPCMessageType in this single logic module. networkextension/XPCServer.swift itself stays excluded
+                // (it would collide with extension/XPCServer.swift's type names + needs no test beyond this decision).
+                "networkextension/NetworkXPCInbound.swift"
             ]
         ),
         .testTarget(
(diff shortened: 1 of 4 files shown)


getvictor:
@coderabbitai full review

PR #332, opened 2026-06-08
Extract a shared XPCEventServer for both extensions

getvictor:
The security and network extensions each carried a near-duplicate XPCServer. That duplication is what let the network extension silently drift: when the hello-ack handshake (issue #178) was added to the security extension's server, the network extension's copy was never updated, so it never answered the agent and delivered no events (fixed in the parent commit). To stop this recurring, unify the two into one shared implementation.

shared/XPCEventServer.swift now owns the listener, peer set, hello-ack handshake, pending buffer, broadcast, peer code-signing, and the dispatchInbound decision -- the proven security-extension implementation, generalized so the inbound application-control handling is an injected `onApplicationControl` hook rather than a hard ApplicationControlStore reference. Each extension owns only a thin instantiation:
- security extension (extension/main.swift): XPCEventServer(... onApplicationControl: { ApplicationControlStore.shared.apply(rawJSON: $0) }).
- network extension (networkextension/XPCServer.swift): XPCEventServer(... nil).

Behavior is preserved for the security extension by construction (the shared class is its former code). The network extension gains the pending buffer and the debug-build ad-hoc peer requirement for free, on top of the handshake. The handshake now lives in exactly one place, so it cannot drift between the two again.

Build integration uses a new `shared/` synchronized file group added to both the `extension` and `networkextension` Xcode targets (no shared framework, per the project's per-slice-entitlements design) and to the EDRExtensionLogic SwiftPM facade. Removed the now-redundant per-extension XPCServer logic + the interim NetworkXPCInbound.swift; the dispatchInbound/PendingBuffer/PeerCodeSigningRequirement unit tests now cover the code both extensions share.

Verified: xcodebuild compiles XPCEventServer.swift into BOTH the extension and networkextension targets (BUILD SUCCEEDED, arm64 + x86_64); swift test (15 logic tests, 0 failures); swiftlint; spectrace 380/380. Runtime sysext verification is deferred (same as the parent handshake fix).

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Refactor**
  * Consolidated XPC server implementation into a shared module, reducing code duplication across extensions.

* **Tests**
  * Removed network extension XPC handshake test.


getvictor:
@coderabbitai full review

PR #333, opened 2026-06-08
Chunk the XPC pending-buffer flush to keep the serial queue responsive

getvictor:
flushPendingTo drained the entire pending buffer (worst case pendingSendCap == 10,000 events) in one synchronous loop on the XPCEventServer's serial queue. While that loop ran, no other queue work could proceed: a second peer's hello or an ERROR disconnect queued behind it had to wait for the whole drain. Each xpc_connection_send_message is a non-blocking enqueue so the loop is not the "hundreds of ms" a blocking IPC round-trip would cost, but a deep drain still monopolises the queue for longer than necessary.

Chunk the flush: drain up-front (clearing the buffer atomically on queue), then send flushChunkSize (256) events per queue iteration, re-dispatching the remainder via queue.async so inbound peer messages interleave between chunks. In-order delivery is preserved (chunks run in index order on the serial queue, in order within each chunk), so the spec contract "delivers every buffered event in the order it was emitted" still holds. A live event broadcast during the flush carries a later timestamp than any buffered event, so the server's timestamp-ordered graph build tolerates the interleave. If the peer disconnected since the previous chunk, stop early rather than sending no-ops at a cancelled connection, consistent with the disconnect-cleanup contract.

Add chunkBoundaries as a pure, total helper so the batching math is unit-tested without an XPC peer (gap-free, non-overlapping, complete, ascending; plus the full-cap 40-chunk case). Both extension targets build.

This addresses the flush half of the Gemini perf review on #332. The other half (offload ApplicationControlStore.apply off the XPC queue) was a false positive and is deliberately not changed: apply already offloads its disk write to persistQueue.async; the only synchronous work on the XPC queue is a JSON decode + dictionary build + atomic in-memory swap, and that swap is intentionally synchronous so AUTH_EXEC sees a pushed policy promptly.

## Checklist

      (flagged below).

## Bug Fixes
* Improved handling of disconnected peers during event transmission, preventing unnecessary send attempts to canceled connections.

## New Features
* Events sent to newly connected peers are now transmitted in optimized batches for improved throughput and responsiveness.

## Tests
* Added comprehensive test cases validating index partitioning behavior, including edge cases, single items, exact multiples, non-power-of-two splits, and large-scale scenarios.

### DIFF
--- a/extension/edr/Tests/EDRExtensionLogicTests/XPCServerLogicTests.swift
+++ b/extension/edr/Tests/EDRExtensionLogicTests/XPCServerLogicTests.swift
@@ -196,46 +196,6 @@ final class XPCServerLogicTests: XCTestCase {
         XCTAssertEqual(buffer.entries, [Data("a".utf8), Data("b".utf8), Data("c".utf8)])
     }
 
-    // The flush of a deep pending buffer is chunked across `queue` iterations (flushChunkSize per chunk) so a worst-case
-    // pendingSendCap drain can't monopolise the serial queue. chunkBoundaries computes the per-chunk index ranges; the
-    // re-dispatch + xpc_connection_send_message loop that consumes them is exercised at the system / VM layer. This test
-    // pins the batching invariants that keep the "delivers every buffered event in the order it was emitted" contract:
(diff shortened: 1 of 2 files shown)


getvictor:
@coderabbitai full review

PR #346, opened 2026-06-12
fix(ingest): batch event inserts instead of one round-trip per event

getvictor:
POST /api/events persisted each event with its own INSERT statement inside one transaction (insertEventsAtOnce looped stmt.ExecContext per event). A ~100-event batch was ~100 sequential MySQL round-trips, which a trace showed as 2331ms of 2369ms spent across 101 sql.stmt.exec spans (detection/graph work was ~38ms). It scales O(batch size) in round-trips, so it worsens as the fleet grows.

Replace the per-event loop with chunked multi-row INSERT IGNORE (mirrors the existing pattern in alerts.go), collapsing N round-trips into one per ~500-row chunk. Idempotent, so withDeadlockRetry still re-runs it safely.

Preserve the ingested_at_ns stamping: fast path (no duplicates) stamps the caller's slice in memory with this batch's time; slow path (INSERT IGNORE dropped a duplicate) reads the persisted ingested_at_ns back via one chunked SELECT and stamps each event from the DB, so a duplicate correctly keeps the time from its first insert. Slice is mutated only after a successful commit so a retried attempt never half-stamps.

No observable behavior change: the idempotency contract (dedup by event_id, existing rows unchanged, HTTP 200) is unchanged; this is the server-event-ingestion "Idempotent submission by event_id" requirement, whose scenarios still hold. Tests added for duplicate-stamping and the chunk boundary; full detection suite incl. integration passes.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Performance**
  * Optimized detection event insertion to use batched database operations, reducing database load and improving throughput.

* **Reliability**
  * Enhanced duplicate detection event handling to correctly preserve original ingestion timestamps when duplicate events are encountered.

### DIFF
--- a/server/detection/internal/mysql/store.go
+++ b/server/detection/internal/mysql/store.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-sql-driver/mysql"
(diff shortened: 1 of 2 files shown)


getvictor:
@coderabbitai full review

PR #347, opened 2026-06-12
refactor(otel): retire bespoke edr.db.query.duration; rely on otelsql…

getvictor:
… native DB metrics

The DB pool is already opened through otelsql (server/bootstrap/db.go), which emits a db.sql.latency histogram for every query plus db.client.connection.* pool metrics with zero per-call-site code (verified flowing in SigNoz). The hand-rolled edr.db.query.duration metric duplicated that, but worse: it needed a manually-maintained bounded-op allowlist (boundedDBOps) policed by an AST-walking static-analyzer test, and it only covered the two call sites someone remembered to instrument (the events insert it was meant to watch wasn't even one of them).

Delete the one-off entirely:
- metrics.go: remove the edr.db.query.duration histogram, its buckets, boundedDBOps + BoundedDBOps(), and ObserveDBQuery.
- detection/api MetricsRecorder: drop the ObserveDBQuery method; intake/handler.go: drop the two call sites + their timing.
- metrics_test.go: drop the histogram assertion and the whole static-analyzer test (TestObserveDBQuery_OperationNamesAreBounded + helpers); integration_test.go: drop the fake's ObserveDBQuery + the assertion.
- observability-instrumentation spec: replace the "DB query latency histogram" requirement with "DB client metrics via standard driver instrumentation" (no bespoke metric); operations.md / install-server.md / ADR-0006 now point at db.sql.latency.

Net: DB latency + pool metrics for EVERY query via the standard instrumentation, sampled-safe (unlike the otelsql spans), and ~150 fewer lines of hand-maintained instrumentation. Per-query timing with request context stays on the otelsql spans. Build, vet, golangci (0 issues), metrics + detection integration tests, spectrace (382/382), prettier/markdownlint/dash-lint all pass.

## Checklist

      (flagged below).


getvictor:
@coderabbitai full review

Issue #360, opened 2026-06-13
Add a retention window for the processes table (mirroring the events retention job)


getvictor:
## Problem

The retention job (`server/detection/internal/pipeline/retention.go`) only deletes from `events` (`EDR_RETENTION_DAYS`, default 30, hourly). The `processes` table has **no retention or pruning whatsoever** — rows accumulate indefinitely. The only process-side background job, `ReconcileStaleProcesses`, just synthesizes a missing `exit_time_ns` for missed-exit rows; it never deletes.

Found on 2026-06-13: a single dev host UUID had accumulated 275K process rows across test sessions. `GET /api/hosts/{host_id}/tree` is bounded (1h window, 2000-row limit) and index-backed (`idx_processes_host_time`), but its "alive at any point in the window" predicate has an unbounded lower time bound (`fork_time_ns < FROM AND exit_time_ns IS NULL`), so EXPLAIN examined ~158K index entries before the LIMIT and the endpoint tripped the 500ms slow-request warning at 1.1s. On a long-lived pilot host the same unbounded growth would surface in production, and restore-from-backup makes it worse.

## Proposal

- Extend the retention runner to prune `processes` on the same `EDR_RETENTION_DAYS` cutoff (by `fork_time_ns`, or by exit time for completed processes), batched like the events delete.
- **Guard against the `alerts.process_id` foreign key** (`fk_alerts_process`, ON DELETE NO ACTION): skip any process referenced by an alert, the same `NOT EXISTS (SELECT 1 FROM alerts ...)` guard used to clear the dev backlog by hand. Without it the batch delete fails on alert-referenced rows.
- Preserve the live snapshot working set (`is_snapshot = TRUE AND exit_time_ns IS NULL`) so an online host's process-tree correlation isn't disrupted.
- Emit the same `edr.retention.rows_deleted` metric (or a process-specific counter) so the prune volume is observable.

## Follow-up that becomes smaller after this

Bounding the `/tree` query's "alive-before-window" lookback (or adding a `(host_id, exit_time_ns)` index for the alive set) is a secondary optimization once the table no longer grows unbounded.

## Context

Surfaced during rc.13 MDM-deploy QA on 2026-06-13; see the `/tree` slow-warning investigation. Spec impact: the retention requirement under `server-event-ingestion` (or wherever events retention is specified) should be extended to cover processes.

getvictor:
[...]

PR #367, opened 2026-06-14
detection: prune the processes table on the retention window (#360)

getvictor:
The retention runner only deleted from events; the processes table had no retention and grew without bound (275K rows on one dev host tripped the /tree slow-request warning at 1.1s). Extend RetentionRunner to also delete completed process rows on the same EDR_RETENTION_DAYS cadence and cutoff.

The prune keys on exit_time_ns, never fork_time_ns: a still-running record (NULL exit, including the live snapshot working set) is never deleted, and a long-running process that only recently exited is retained for the full window measured from its exit. Stale missed-exit rows are force-closed by the freshness-TTL reconciler (#6, default-on) and swept here once their synthesized exit ages past the window: TTL marks dead, retention deletes old-dead. The alerts.process_id FK is ON DELETE RESTRICT, so the batch DELETE skips alert-referenced rows via NOT EXISTS.

- retention.go: batched processes DELETE, shared pruneBatched helper.
- migration 00002: idx_processes_exit_time backing the ordered range delete.
- metrics: dedicated edr.retention.processes.rows_deleted counter.
- spec: server-process-graph-builder requirement + 3 scenarios; openspec change proposal add-processes-retention.
- integration test asserts prune/keep across completed, live (snapshot and non-snapshot), and alert-referenced rows.

## Checklist

      (flagged below).

## Release Notes

* **New Features**
  * Automated cleanup of old completed processes is now enabled, using the same retention window as event retention (EDR_RETENTION_DAYS).
  * Active processes and those referenced by alerts are automatically protected from deletion.
  * New observability metric added to monitor process retention activity.

### DIFF
--- a/openspec/changes/add-processes-retention/proposal.md
+++ b/openspec/changes/add-processes-retention/proposal.md
@@ -1,0 +1,19 @@
+# Processes retention window
+
+## Why
+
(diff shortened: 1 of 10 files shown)


getvictor:
@coderabbitai full review

Issue #379, opened 2026-06-15
Detection pipeline wedges fleet-wide on uid/gid overflow (processes.uid is signed INT)


getvictor:
## Summary

The detection pipeline stalls fleet-wide when a process event carries a macOS uid/gid ≥ 2³¹. `processes.uid`/`gid` are declared signed `INT` (max 2,147,483,647); macOS `uid_t`/`gid_t` are unsigned 32-bit. `nobody` is `4294967294` (-2 as uint32) and `KAUTH_UID_NONE` is `4294967295` (-1), both of which overflow signed `INT` and fail the insert with **MySQL Error 1264 (22003) "Out of range value for column 'uid'"**.

## Severity: critical (sev-1) for the detection plane

It is not a one-row drop. The graph builder fails the whole batch on any per-event error (`builder.go:124-144`), and the processor then unclaims the batch and never marks it processed (`processor.go:82-88`), so the poison event stays the oldest unprocessed row and is re-fetched + re-failed every cycle. Detection rule evaluation runs only after the batch succeeds, so it never runs → **no new process-graph materialization and no new alerts, fleet-wide, with no self-recovery.** Triggered by a routine macOS uid.

On-device app-control blocking (cached policy in the system extension) and raw event ingestion are unaffected.

## Evidence (prod, `prod-render`, v0.1.1, SigNoz)

- 6h window: `event processing failed` ×55,920 with `err: insert process: Error 1264 (22003): Out of range value for column 'uid' at row 1`, paired ~1:1 with `graph builder failure, will retry batch` ×55,935 (~2.6 retries/sec on a single poisoned batch). Example event type `exec`, event_id `9F698518-5232-4DA3-BA8F-4BE30FB8BE7E`.

## Root cause

- `server/detection/migrations/00001_initial.sql:38-39` — `uid INT, gid INT` (signed).
- `server/detection/internal/graph/builder.go:124-144` — any per-event failure fails the whole batch.
- `server/detection/internal/pipeline/processor.go:82-88` — batch error → unclaim all, never `MarkProcessed` → permanent poison-pill loop.

## Fix

1. Migration: `ALTER TABLE processes MODIFY uid INT UNSIGNED, MODIFY gid INT UNSIGNED;` (forward-only; stored values are all 0..2³¹-1, lossless).
[...]

PR #381, opened 2026-06-15
fix(detection): widen processes uid/gid to unsigned, isolate poison e…

getvictor:
…vents

processes.uid/gid were signed INT (max 2147483647), but macOS uid_t/gid_t are unsigned 32-bit. nobody (4294967294) and KAUTH_UID_NONE (4294967295) overflowed and failed the insert with MySQL error 1264. The builder failed the whole batch on any per-event error and the processor re-fetched it forever, so one such event wedged the detection pipeline fleet-wide (no new process graph or alerts), with no self-recovery. See #379.

- migration 00003: ALTER processes MODIFY uid/gid INT UNSIGNED (lossless; stored values are all within the old signed range).
- mysql.IsPermanentDataError: classify data-integrity error numbers (1264/1265/1292/1366/1406/3819) that recur on every retry.
- builder.ProcessBatch: drop an event that fails permanently (log 'event dropped: permanent processing error') so the batch advances; fail the batch only for transient errors so the processor retries.
- tests: classifier unit test; integration tests proving a nobody-uid exec persists and a poison event is dropped while the batch still advances.
- openspec change delta for the gated migrations path.

## Checklist

      (flagged below).

### DIFF
--- a/openspec/changes/fix-process-uid-gid-overflow/proposal.md
+++ b/openspec/changes/fix-process-uid-gid-overflow/proposal.md
@@ -1,0 +1,17 @@
+# Fix process uid/gid overflow wedging the detection pipeline
+
+## Why
+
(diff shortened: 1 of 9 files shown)


getvictor:
@coderabbitai full review

PR #385, opened 2026-06-15
Verify host tokens with keyed HMAC, derive keys from a root secret

getvictor:
Host-token verification recomputed argon2id (t=3, m=64 MiB, p=4) fresh on every authenticated agent request, with no cache. Production telemetry (prod-render) showed both agent endpoints at a ~235ms p50 dominated by that hash, tripping the 500ms slow-request log under concurrency. A high-entropy random bearer token does not need a memory-hard password KDF; a fast keyed hash gives the same resistance to offline recovery.

Changes:
- Replace argon2id with HMAC-SHA256(pepper, token), constant-time compare. Drop the per-row salt columns (migration 00002).
- Add internal/keyring: HKDF-SHA256 key separation from one root secret.
- Fold EDR_SESSION_SIGNING_KEY into EDR_SECRET_KEY: the host-token pepper and the pre-auth cookie signing key both derive from it. EDR_SECRET_KEY is now required on every boot; EDR_SESSION_SIGNING_KEY is removed.
- Wire derived keys through cmd/main into the endpoint and identity contexts.
- Update env surfaces (Taskfile, render.yaml, compose, scripts) and docs.

Breaking change: existing enrollments hold argon2id hashes that no longer verify, so every host re-enrolls on its next request via the revocation path. No token re-hash migration.

openspec: host-token-fast-hash (agent-enrollment + server-identity-authentication deltas)

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **New Features**
  * Enhanced host-token verification performance through improved cryptographic operations

* **Chores**
  * Required environment configuration change: `EDR_SECRET_KEY` now replaces `EDR_SESSION_SIGNING_KEY`
  * **Breaking Change:** All existing host tokens and operator sessions are invalidated after deployment and require re-enrollment and re-authentication


getvictor:
@coderabbitai full review

PR #402, opened 2026-06-17
fix(app-control): re-sync after a server policy-version regression (#…

getvictor:
…322)

The extension gated app-control snapshots on a monotonic per-policy policy_version, equating "stale" with "lower version". policy_version lives only in the DB, so a restore-from-backup (the documented rollback path) or reset regresses it below what hosts have persisted, and the extension then silently ignores every subsequent update for that policy: new rules never enforce, removed rules keep blocking, with only an Info log on the host.

Carry the policy's updated_at as policy_epoch (Unix microseconds) alongside policy_version. Wall-clock survives a restore (the next mutation stamps a fresh updated_at strictly greater than any pre-restore epoch), so the gate now accepts a snapshot when EITHER version OR epoch advanced, and rejects (no-op) only when both are <= the active snapshot's. This preserves replay/out-of-order protection (older on both axes) while re-syncing after a regression. A version regression accepted via the epoch axis is logged at error and emitted as a new application_control_resync host event so the regression is observable. Hosts already frozen before this change self-heal on upgrade (persisted epoch 0 -> N).

- server/rules/api: PolicyEpoch on SetApplicationControlPayload, marshalled from policy.UpdatedAt.UnixMicro() (0 sentinel when unset); wire tests + PBT.
- extension: epoch field on snapshot/document, new gate, resync event + reporter wiring; ApplicationControlStore unit tests for the epoch/regression paths and the resync event.
- schema/events.json: application_control_resync event + payload.
- test/integration: real-MySQL regression test asserting epoch advances on a version regression.
- openspec change fix-app-control-version-regression-resync.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of application-control policy version regressions (e.g., after database restore) by introducing an epoch-based synchronization mechanism that enables automatic recovery instead of policy freezing.

* **New Features**
  * Added resynchronization event tracking to provide visibility when the system recovers from policy version regressions.

* **Documentation**
  * Updated specifications and task documentation to reflect epoch-based policy gating logic and resync event requirements.

* **Tests**
  * Added comprehensive test coverage for epoch-based policy acceptance and version regression handling.


getvictor:
@coderabbitai full review

PR #431, opened 2026-06-18
fix(detection): select re-exec generation by event time within a pidv…

getvictor:
…ersion (#428 review)

A same-PID re-exec chain shares one kernel PID generation (execve keeps the proc), so (host_id, pid, pidversion) matches every generation in the chain, not one. GetProcessByPIDVersion returned a single row by 'live, else newest id' and resolveFlowProcess ignored the event timestamp on the identity path, so a flow opened by an earlier generation (e.g. curl) that then re-execs (e.g. cleanup) was mis-attributed to the live/newest generation. Two PR #428 reviewers (Gemini, Qodo) flagged it.

GetProcessByPIDVersion now takes an atNs anchor. A single identity match is returned regardless of atNs (PID reuse stays skew-immune); when the identity matches multiple generations it selects the one that was the running image at atNs, bracketing on COALESCE(exec_time_ns, fork_time_ns) <= atNs < exit_time_ns. A re-exec chain preserves the original fork_time_ns on every generation, so exec_time_ns is the boundary that orders them. The no-pidversion and identity-miss paths fall back to the GetProcessByPID time window unchanged.

Ships the openspec delta (flow-reexec-correlation) updating the 'linked at event time' requirement with the within-identity timestamp-selection rule and a re-exec scenario, plus a real-MySQL test asserting a flow during an earlier generation links to it, not the live one.

## Checklist

      (flagged below).


getvictor:
@coderabbitai full review

Issue #408, opened 2026-06-17
Reduce events-table disk growth (drop heartbeat rows, coalesce network/DNS, index/PK diet) and evaluate ClickHouse as the event store


getvictor:
## Summary

The `events` table dominates server disk and grows ~8-10 GB/day/host (observed on the Render pilot; corroborated by the `render.yaml` comment "10G fills in ~a night ... events are ~1.3KB each"). Diagnostics from a few hours of one host's data:

| event_type | rows | % of rows | payload_mb |
|---|---|---|---|
| snapshot_heartbeat | 218,389 | 22% | 3 |
| network_connect | 193,438 | 19% | 57 |
| exec | 155,564 | 16% | 79 |
| dns_query | 148,833 | 15% | 36 |
| fork | 139,864 | 14% | 5 |
| exit | 139,682 | 14% | 4 |

```
table     gb    data_gb  index_gb
events    1.31  0.52     0.79
processes 0.17  0.11     0.06
```

Key findings:

- **Disk tracks row count, not payload bytes.** Each `events` row is ~1.3 KB on disk: 6 secondary indexes plus the `VARCHAR(255)` `event_id` primary key copied into every one of them. On the events table, index (0.79 GB) > data (0.52 GB).
- **`snapshot_heartbeat` is 22% of rows for ~zero forensic value.** It is an agent liveness ping (one per still-alive snapshot PID every `EDR_PROCESS_RECONCILE_INTERVAL`, default 60s; ~900 PIDs on a normal macOS host). Its only server-side effect is bumping `processes.last_seen_ns`; the detection engine already filters it (`server/detection/internal/engine/filter.go`). Yet it is INSERTed into the heaviest table, fully indexed, and retained for the whole `EDR_RETENTION_DAYS` window.
- **`network_connect` + `dns_query` = 34% of rows**, much of it repetitive (same process to same destination; the same domain resolved repeatedly).

### 1. Stop persisting `snapshot_heartbeat` rows in `events` (highest impact, lowest risk)

Process the heartbeat for its `processes.last_seen_ns` bump, then drop it instead of writing a retained `events` row. Removes ~22% of event rows and their index overhead.

[...]

Issue #405, opened 2026-06-17
Optimize agent->server telemetry upload: compression, persistent connections, modern transport


getvictor:
## Summary

Optimize the agent -> server telemetry upload path (`POST /api/events`) for bandwidth and connection efficiency. Today the uploader posts raw, uncompressed JSON batches over a fresh request per tick. On the insert-heavy ingest workload (events are ~1.3 KB each, batches up to 100) this is wasteful on the wire and on per-request overhead. Explore modern best practices, not just compression.

This is a performance/efficiency issue. It is **not** the fix for the WAF false-positive incident (that is architectural: exempt the authenticated agent-ingest routes from edge signature inspection). Compression does not reliably evade a WAF that decompresses, so the two are tracked separately.

## Implementation status (updated 2026-06-22)

Partially implemented. The connection-efficiency cluster (persistent connections + HTTP/2) has landed; payload-size work (compression, binary format) and the push-based control channel are still open.

- [ ] **Option 1: Request-body compression (gzip/zstd).** OPEN. This is the headline remaining win. The uploader sends raw JSON with `Content-Type: application/json` and no `Content-Encoding` (`agent/uploader/uploader.go:239,470,474`); the intake handler does not decompress and enforces only a wire-byte cap, not a decompressed-size cap (`server/detection/internal/intake/handler.go` `readBodyWithCap`). Both ends and the zip-bomb guard remain to do.
- [x] **Option 2: Persistent connections / keep-alive.** DONE. A single cloned `http.DefaultTransport` is built once and shared by the uploader and commander, preserving keep-alive, proxy support, and hardened dial/idle timeouts (`agent/cmd/fleet-edr-agent/main.go` `newAgentHTTPClient`, ~L260).
- [x] **Option 3: HTTP/2.** DONE (#454). `http2.ConfigureTransports` negotiates h2 over the pinned TLS config with keep-alive PINGs (`ReadIdleTimeout`/`PingTimeout`) to drop half-open links; falls back to HTTP/1.1 keep-alive on failure. HPACK amortizes the per-request bearer/header cost called out in the original "Current state". OTel propagation is preserved (`otelhttp.NewTransport` still wraps the base transport).
- **Option 4: Streaming / long-lived channel.** MOVED to #477. It is a control-path concern (replace `GET /api/commands` polling with server push), not telemetry-upload, and overlaps the scale-to-100k control-channel design. Tracked separately there.
[...]

PR #432, opened 2026-06-18
feat(events): cut events-table disk growth (#408)

getvictor:
Three independent reductions to events-table disk, each shipped with an openspec change delta:

1. Drop snapshot_heartbeat persistence (~22% of rows). The ingest path now applies the heartbeat's only side effect (the processes.last_seen_ns freshness bump that exempts a live snapshot row from the 6h TTL) directly at ingest and drops the event instead of writing a retained row. Host liveness (UpsertHosts) and the accepted count are unchanged. New counter edr.ingest.heartbeats_dropped. The graph-builder + engine-filter heartbeat paths remain as a defensive path for pre-upgrade rows.

2. Coalesce repetitive network_connect / dns_query at the agent before enqueue. New agent/coalesce package buffers these two types in a bounded window (EDR_NETWORK_COALESCE_WINDOW, default 10s, 0 disables) and merges events sharing an identity key into one representative + coalesced_count, preserving the earliest timestamp (so the 30s DNS->connect beacon window is never shortened), last_timestamp_ns, and the union of DNS response_addresses. Every other event passes through immediately; the buffer flushes on shutdown.

3. Events index/PK diet. Drop the two redundant secondary indexes (idx_events_host_id subsumed by the host_id composite; idx_events_type serves no query) and replace the 36-char UUID primary key with a BIGINT AUTO_INCREMENT surrogate, keeping event_id as a UNIQUE key. event_id stays the logical identity on the wire and in every query; INSERT IGNORE dedup and the alert_events FK are unchanged. Each secondary-index leaf now carries an 8-byte surrogate instead of the UUID.

Verified live on dev:server + edr-dev: migrations applied to the populated dev DB (PK rebuilt to the surrogate, redundant indexes gone); heartbeats no longer persisted while last_seen still bumps and edr.ingest.heartbeats_dropped flows to SigNoz (dev-local); the real coalescer collapsed 30 repetitive network/DNS events to 5 representative rows (counts + DNS address union preserved, distinct 5-tuples kept separate). Detection efficacy corpus, openspec validate --all --strict, and spectrace all green.

Not in scope: ClickHouse evaluation (deferred to post-deploy numbers); upload-bandwidth optimization (#405).

## Checklist

      (flagged below).

## Release Notes

* **New Features**
  * Agent-side coalescing of repetitive network/DNS telemetry events within a configurable time window (default 10 seconds; configurable via `EDR_NETWORK_COALESCE_WINDOW`, set to 0 to disable)
  * Heartbeat freshness applied at ingestion without persisting events to storage

* **Performance**
  * Database schema optimized with surrogate primary key and redundant index removal for improved query performance

* **Documentation**
  * Operations guide updated with event volume reduction guidance and observability metrics to monitor effectiveness


getvictor:
@coderabbitai full review

Issue #253, opened 2026-05-23
agent-event-uploader: permanent 4xx is not retried up to max + no audit log


getvictor:
## Spec / impl drift

`openspec/specs/agent-event-uploader/spec.md:119-129` (`permanent-client-errors-are-not-infinitely-retained`):

> **GIVEN** the server returns a non-401 4xx for a batch on every attempt
> **WHEN** the configured maximum retry budget is exhausted
> **THEN** the uploader records an audit log entry identifying the failure
> **AND** the batch is not retransmitted indefinitely

`agent/uploader/uploader.go` `uploadWithRetry`:

```go
// Don't retry client errors (4xx) — only server/network errors are retryable.
var clientErr *clientError
if errors.As(err, &clientErr) {
    return clientErr
}
```

Behavior gap vs spec:

1. The uploader treats ALL 4xx (including non-401) as immediate-non-retryable on the first attempt; the
   configured retry budget is never consumed for 4xx. Spec says "after the configured maximum" — implies
   retry up to max, then stop.
2. There is no audit log call for non-401 4xx. Only `u.logger.ErrorContext(ctx, "uploader upload failed", ...)`
   fires, which is a regular log line, not an audit entry.
3. The batch stays in the queue (`drainOnce` returns the error, MarkUploaded is never called). Next cycle
   will dequeue the same batch and try again, indefinitely. Spec says "not retransmitted indefinitely."

## Fix shape

Two distinct decisions, in order:

1. **Retry-or-not on 4xx**: keep the current "no retry on 4xx" stance (server isn't going to flip its
   mind in 100ms) but document this explicitly in the spec, OR implement the spec's "retry up to max
   then stop" semantics. The current behavior is arguably correct for client errors, but the spec needs
   to match either way.
2. **Indefinite retransmission**: after the first 4xx (or after max retries if we go with the spec
   shape), the batch should be quarantined — either marked uploaded with a `failed_at` audit row, or
   moved to a poison-message table for operator review. Today it just sits in the queue forever, which
   means a single malformed event blocks every subsequent batch.

## Why deferred from #248's stack

This is real behavior change touching event-delivery semantics. Out of scope for a marker rollout PR.
The marker for this scenario will land once the spec + impl agree.

getvictor:
Closing. Resolved during the M13 marker stack:

[...]

PR #449, opened 2026-06-19
fix(agent): keep good telemetry queued on a blanket endpoint rejection (#398)

### LINKED ISSUE #398: Agent uploader drops good telemetry on a sustained 403 (treats edge/infra 4xx as a poison-event)
## Summary

The agent uploader treats a persistent HTTP **403** on `POST /api/events` the same as a malformed-payload 4xx: after `ClientErrorQuarantineThreshold` consecutive failed drain ticks (default 10) it marks the batch `uploaded=1` and drops it (#253 quarantine). But a 403 on the ingest route is **never** the EDR server's own response, so it is always an *infrastructure* signal (edge/proxy/WAF/access layer, or a wrong/unhealthy origin), not a "this batch's content is bad" signal. The result is silent, bounded **data loss** of good telemetry during an outage that a 401-style "back off and keep the queue" would have preserved.

## Where

`agent/uploader/uploader.go` `handleUploadErr` routes upload failures:

- **401** -> `OnAuthFail` (re-enroll), batch stays queued. Recoverable.
- **413** -> bisect/split, single-event drop; does NOT consume the quarantine budget.
- **5xx / network** -> log + return, batch stays queued and retries. Recoverable.
- **non-401 4xx (incl. 403)** -> `recordClientErrorAndAudit` -> after `quarantineThreshold` ticks, `RecordClientError` sets `uploaded=1` in `agent/queue/queue.go` and the row stops being dequeued. **Dropped.**

So 403 falls into the quarantine-and-drop bucket alongside 400 (`invalid_json`, `host_id_mismatch`, `missing_fields_at_<i>`).

## Why 403 is different from the other 4xx here

The ingest path (`POST /api/events`) is wrapped by the host-token middleware only (`server/cmd/fleet-edr-server/main.go`, `server/endpoint/internal/middleware/hosttoken.go`): it returns **401** (`invalid_token`/`missing_bearer`), **503** (`verifier_unavailable`), or passes to the detection intake handler (`server/detection/internal/intake/handler.go`), which returns **200 / 400 / 413** only. There is no 403 anywhere on the agent path. Therefore a 403 the agent sees is always injected by something in front of the server.

Treating it as a permanent per-batch content error is the wrong recovery: the content is fine, the endpoint is rejecting everything.

## Observed (real incident, 2026-06-16)

Fresh Render deploy; `EDR_SERVER_URL=https://<svc>.onrender.com` returned 403 at the edge (web service not healthy behind Render). Agent log:

```
ERROR uploader upload failed ... err="server returned 403" batch_size=...
ERROR uploader quarantined events after persistent client errors ... status_code=403 threshold=10
```

Every batch eventually burned its 10-tick budget and was dropped while the endpoint was unreachable.

## Proposed fix

Distinguish "this batch's content is bad" from "the endpoint is rejecting everything":

- Treat **403** like a retryable/transient condition (same bucket as 5xx/network): keep the batch queued, back off, resume when the endpoint recovers. The EDR app never legitimately 403s this route, so there is no risk of an infinitely-stuck poison batch from a real 403.
- Optionally generalize: if *every* batch is failing with the same status (endpoint-level rejection) rather than one specific batch, that is an endpoint signal -> back off, do not quarantine. The quarantine budget should be reserved for a batch that fails while *other* batches succeed (the true poison-event shape).
- Emit a distinct, loud signal for the "endpoint rejecting all uploads" state (a dedicated WARN + counter) so an operator sees "server unreachable / misconfigured" rather than a quiet drop.

## Acceptance

- A sustained 403 (or other blanket edge rejection) does NOT drop events; the queue is preserved and uploads resume automatically when the endpoint returns 200 (bounded only by `EDR_AGENT_QUEUE_MAX_BYTES`).
- 400-class content errors (`invalid_json`, `host_id_mismatch`, `missing_fields_at_<i>`) still quarantine as today (genuine poison events).
- A distinct log line / metric marks the "endpoint rejecting uploads" state.
- Tests in `agent/uploader/uploader_test.go` cover: sustained 403 preserves the queue; a single poison 400 still quarantines while sibling batches deliver.

## Notes

Documentation already records that releases predating this carry the old behavior; no spec/runtime contract pins 403 -> quarantine, so this is a safe behavior change. Pair with a delta under `openspec/changes/` since it changes observable agent persistence behavior.

getvictor:
## Summary

Fixes #398. The agent uploader treated any non-401 4xx on `POST /api/events` as a per-batch content error: after `ClientErrorQuarantineThreshold` consecutive failed drain ticks (default 10) it sealed the batch (`uploaded=1`) and dropped it (#253 quarantine). But the EDR ingest route only ever emits **200/400/413** (intake) and **401/503** (host-token middleware), so a **403** (or 404/429/...) is always an edge/proxy/WAF or unhealthy-origin signal, never "this batch is bad". A real Render incident (2026-06-16) dropped good telemetry this way.

## What changed

- **Quarantine is now reserved for HTTP 400** (the only content-poison status the ingest contract emits). A persistent 400 still seals the batch after the threshold, exactly as before.
- **Every other non-2xx the route never emits itself (any 4xx except 400/401/413) becomes a transient `endpointRejectedError`**: the batch stays queued (`uploaded=0`), the uploader backs off per drain tick, and uploads resume automatically on 2xx. Retention during the rejection window is bounded only by the existing `EDR_AGENT_QUEUE_MAX_BYTES` lossy cap, not by the quarantine budget.
- **A distinct, loud signal** marks the state: a WARN (`audit=uploader.endpoint_rejecting`, with `status_code`) plus a new OTel counter `edr.agent.uploader.endpoint_rejected` (by `status_code`).
- 401 (re-enroll) and 413 (split-and-retry) are unchanged.

## Spec

`openspec/changes/uploader-edge-rejection-no-quarantine/` narrows the "Permanent client errors" requirement to 400 and adds "Blanket endpoint rejections keep the queue". `openspec validate --strict` passes.

## Tests

`agent/uploader/uploader_test.go`: sustained 403 preserves the queue and resumes on 200; a persistent 400 still quarantines while a sibling batch delivers; 404/429 are kept queued. `-race` green; `task lint:go` clean; `spectrace check --strict` 419/419.

## Manual verification (live)

Ran the real patched agent binary against the live dev server through a toggleable 403 edge, exporting to SigNoz (`deployment.environment=dev-local`):

- **Sustained 403: 70 drain ticks, 0 events dropped, 0 quarantine lines** (old behavior dropped at tick 10). Each tick logged `uploader endpoint rejecting uploads`.
- The new `edr.agent.uploader.endpoint_rejected` counter reached SigNoz.
- **Recovery (flip to 200): all 200 preserved events delivered** to the dev server (`SELECT COUNT(*) ... = 200`), agent went quiet, no quarantine/errors.

### DIFF
--- a/agent/metrics/metrics.go
+++ b/agent/metrics/metrics.go
@@ -41,6 +41,7 @@ type QueueDepthSource interface {
 type Recorder struct {
 	queueDropped          metric.Int64Counter
 	eventsDroppedTooLarge metric.Int64Counter
+	endpointRejected      metric.Int64Counter
 	queueDepth            metric.Int64ObservableGauge
 }
 
@@ -64,6 +65,11 @@ func NewWithMeter(depthSrc QueueDepthSource, m metric.Meter) *Recorder {
 		"edr.agent.uploader.events_dropped_too_large",
 		metric.WithDescription("Events the uploader dropped after a single-event batch was rejected by the server with HTTP 413 `body_too_large`. A non-zero rate indicates an agent producing events larger than the server's per-request cap."),
 		metric.WithUnit(unitEvent),
+	)
+	r.endpointRejected, _ = m.Int64Counter(
+		"edr.agent.uploader.endpoint_rejected",
+		metric.WithDescription("Drain ticks on which the ingest endpoint returned a blanket rejection (a 4xx the EDR server never emits itself, e.g. 403/404/429), by `status_code`. Non-zero means the agent is queueing telemetry it cannot deliver: the endpoint URL is wrong, the origin is unhealthy, or an edge/WAF is blocking. The batch is kept queued, not dropped."),
+		metric.WithUnit("{rejection}"),
 	)
 	if depthSrc != nil {
 		r.queueDepth, _ = m.Int64ObservableGauge(
@@ -105,4 +111,14 @@ func (r *Recorder) EventsDroppedTooLarge(ctx context.Context, n int64) {
 		return
 	}
 	r.eventsDroppedTooLarge.Add(ctx, n)
+}
+
+// UploadRejected increments the endpoint-rejected counter once per drain tick on which the ingest endpoint returned a blanket
+// rejection (a 4xx the EDR server never emits itself: 403/404/429/...). The batch is kept queued, not dropped; this counter is the
+// operator signal that the endpoint is unreachable or misconfigured (#398). Nil-safe.
+func (r *Recorder) UploadRejected(ctx context.Context, statusCode int) {
+	if r == nil || r.endpointRejected == nil {
+		return
+	}
+	r.endpointRejected.Add(ctx, 1, metric.WithAttributes(attribute.Int("status_code", statusCode)))
 }
--- a/agent/uploader/uploader.go
+++ b/agent/uploader/uploader.go
@@ -32,8 +32,9 @@ const (
 	// up and falls through to the next drain tick.
 	defaultMaxRetries = 5
 
-	// defaultClientErrorQuarantineThreshold caps the number of consecutive drain ticks a row's batch may return a non-401 4xx
-	// before the row is quarantined (uploaded=1, removed from the dequeue set, audit log emitted). 10 matches the spec's
+	// defaultClientErrorQuarantineThreshold caps the number of consecutive drain ticks a row's batch may return an HTTP 400 (the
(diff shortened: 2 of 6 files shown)


getvictor:
@coderabbitai review

getvictor:
@coderabbitai review

PR #454, opened 2026-06-19
feat(auth): self-validating signed host tokens with agent-pull refresh

getvictor:
Replace opaque-bearer + per-request DB verify with self-validating signed tokens.
A token (v1.<claims>.<hmac>, claims hid/ep/iat/exp/kid) is signed with a key derived
from EDR_SECRET_KEY via HKDF; VerifyToken is a local signature + expiry check plus an
in-memory revocation snapshot, with no database lookup on the agent hot path.

- signedtoken: mint/verify (HMAC-SHA256, constant-time, kid + expiry checks) + PBT.
- revocation: per-replica snapshot (token_epoch + revoked_at), refreshed every 5s,
  loaded before serving; OTel size/age/refresh-failure instruments. Migration 00003
  adds token_epoch.
- service: signed verify/mint, RefreshToken (agent-pull), RotateToken bumps the epoch;
  verify-time auto-rotation + grace window removed. EnrollResponse carries expires_at.
- POST /api/token/refresh behind the host-token middleware; mounted in both binaries,
  which also run the snapshot refresh loop. EDR_HOST_TOKEN_LIFETIME default now 60m.
- agent: persists/parses expires_at, proactive refresh loop (~2/3 TTL), HTTP/2 keep-alive
  PINGs on the shared transport.

Hard cutover: pre-existing opaque tokens 401 and re-enroll. Revocation is eventually
consistent, bounded by the snapshot refresh interval. OpenSpec change + spec delta in
openspec/changes/signed-host-tokens. Verified on the edr-dev VM + dev server + SigNoz.

Claude-Session: https://claude.ai/code/session_01WiSixU4Nhfvz9cxvotjL5L

## Release Notes

* **New Features**
  * Agents now automatically refresh authentication tokens before expiry to prevent connection lapses.
  * Added HTTP/2 support with keep-alive optimization to the agent's outbound connections.
  * Introduced a new operator-driven token rotation mechanism using epoch-based invalidation.

* **Changes**
  * Reduced default token lifetime from 24 hours to 60 minutes.
  * Tokens are now self-validating with local verification instead of database lookups.
  * Token revocation now enforced via in-memory snapshot with periodic updates.


getvictor:
@coderabbitai full review

Issue #375, opened 2026-06-15
UI to configure SSO/OIDC (replace EDR_OIDC_* env-var-only setup)


getvictor:
## Summary

SSO (OIDC) is configurable only through environment variables today. We need an in-product admin UI to configure SSO, so an operator can stand up and edit the IdP integration without editing env vars and restarting the server.

## Today's state (env-var only)

OIDC is populated entirely from `EDR_OIDC_*` env vars, validated as a coherent set at boot (`server/config/config.go:464` onward):

- `EDR_OIDC_ISSUER`, `EDR_OIDC_CLIENT_ID`, `EDR_OIDC_CLIENT_SECRET`, `EDR_OIDC_REDIRECT_URL`
- `EDR_OIDC_STATE_COOKIE_TTL`, `EDR_OIDC_SCOPES`, `EDR_OIDC_ALLOW_JIT_PROVISIONING`, `EDR_OIDC_DEFAULT_ROLE`
- `EDR_AUTH_ALLOW_NO_OIDC=1` to run break-glass-only.

Every change requires editing deployment env and restarting. `docs/okta-setup.md` documents the manual flow.

## Why it matters

Every commercial EDR / security console ships SSO setup as an in-product admin screen (CrowdStrike, SentinelOne, Defender, Elastic). Pilot customers expect to point the product at their IdP (issuer, client id/secret, redirect URL, scopes, JIT default role) from a settings page, test the connection, and save, without a DBA/ops ticket.

## Scope

- Admin settings page to view + edit the OIDC config: issuer, client id, client secret (write-only / masked), redirect URL, scopes, JIT toggle, default role.
- A "test connection" affordance (discovery doc fetch + a dry-run of the redirect) before saving.
- Persist config durably (MySQL) so it survives restarts and is consistent across replicas (stateless-server / ADR-0010): env vars become the bootstrap/override, the stored config is the runtime source of truth. Decide precedence (env override vs stored) explicitly.
- Gate behind the existing authz chokepoint (admin / a new `sso.manage` action). Audit every mutation.

## Interplay with existing issues

- #136 (OIDC group-claim -> role mapping) and #135 (role-management UI) are the RBAC half. This issue is the IdP-connection half. They share the identity context but are separable.
- #66 (Add SSO + RBAC, closed) shipped the wave-1 env-var-only OIDC. This is the wave-2 "configure it from the product" follow-up.

## Acceptance

[...]

PR #473, opened 2026-06-21
fix(server): derive the session-protected route allowlist (#463)

### LINKED ISSUE #463: Guard against authed routes missing from the buildMux session-protected allowlist (recurring 302→SPA bug)
## Summary

`buildMux` in `server/cmd/fleet-edr-server/main.go` mounts the session-protected `apiMux` onto the outer router via an **explicit, hand-maintained allowlist** of authed paths (the `for _, p := range []string{...}` block). Every context registers its authed routes on `apiMux` via `RegisterAuthedRoutes`, but a route only actually serves traffic if its exact pattern is *also* listed in that slice. A route registered on `apiMux` but missing from the slice falls through to the `/` SPA catchall and returns **302 → /ui/**; the UI's `fetch` then parses the SPA's `index.html` as JSON and surfaces `Unexpected token '<' … is not valid JSON`.

This has now bitten twice:

- **PR #158** — app-control list endpoint returned the SPA shell (noted in the comment in that block).
- **PR #375 / #461** — `GET /api/settings/sso` 302'd to `/ui/`; only caught by live dev-server QA, after the unit + integration tests passed. The handler/integration tests mount `RegisterAuthedRoutes` on a bare mux, so they bypass the allowlist entirely and cannot catch a missing entry.

## Why it keeps happening

The allowlist duplicates routing knowledge that already lives in each context's `RegisterAuthedRoutes`. The two can silently drift: nothing fails the build, no unit test exercises the composed `buildMux`, and the only symptom is a runtime 302 that looks like a UI bug.

## Proposed fix

A regression guard that asserts every authed route a context registers is reachable through the *composed* `buildMux` (i.e. returns 401/JSON when unauthenticated, **not** 302 → /ui/). Options, roughly in order of preference:

1. **Derive the allowlist instead of hand-maintaining it.** Have each context expose the list of authed patterns it registers (or capture them via a recording `*http.ServeMux` during `RegisterAuthedRoutes`), and have `buildMux` mount exactly those on `sessionProtected`. This removes the duplication so drift is impossible.
2. **If the allowlist must stay explicit, add a `buildMux`-level test** that, for every authed pattern each context registers, issues an unauthenticated request through the fully composed mux and asserts the response is the session middleware's `401 application/json` rather than a `302` to `/ui/`. This catches a missing entry the moment a new authed route is added.

## Acceptance

- Adding a new authed route without updating the `main.go` allowlist fails a test (or is impossible because the list is derived), rather than shipping a route that 302s to the SPA.
- The existing routes (app-control surface, audit, session, reauth, SSO settings) are covered by the guard.

## Context

Found during the SSO/OIDC config UI work (#375). The missing SSO entries were fixed in that PR's branch (commit adding `GET/PUT /api/settings/sso` + `POST /api/settings/sso/test-connection` to the allowlist); this issue tracks the durable guard so it does not recur a third time.

getvictor:
## Why

`buildMux` mounted the session-protected API via a hand-maintained allowlist slice (`for _, p := range []string{...}` in `server/cmd/fleet-edr-server/main.go`) that had to stay in sync with each context's `RegisterAuthedRoutes`. The two drifted twice (#158 app-control, #375 SSO): a route registered on the API mux but missing from the slice fell through to the `/` SPA catch-all and served `index.html`, which the UI's `fetch` parsed as JSON (`Unexpected token '<'`). Unit/integration tests mount `RegisterAuthedRoutes` on a bare mux and bypass the composed allowlist, so only live QA caught it. Closes #463.

## What changed

`http.ServeMux` can't enumerate its registered patterns, so the only way to stop the drift is to capture patterns at registration time.

- Add `httpserver.Router` (`Handle` + `HandleFunc`, satisfied by `*http.ServeMux`) and `httpserver.RecordingRouter` (forwards to an inner router while recording patterns).
- Change the authed registration methods (the 5 contexts' `RegisterAuthedRoutes` + the operator handlers' `RegisterRoutes` they delegate to) to accept `httpserver.Router`. `*http.ServeMux` satisfies it, so every call site compiles unchanged and the method bodies are untouched.
- `registerSessionRoutes` registers every context through one `RecordingRouter` (via the extracted, testable `mountAuthed`) and mounts exactly the recorded patterns on the session-protected boundary. The hand-maintained slice is deleted. A registered authed route is now allowlisted automatically, so the drift is structurally impossible rather than guarded after the fact.

**No observable change to any existing route** — same routes, same responses; only future drift is prevented.

## Out of scope

The agent-token allowlist in `registerHostRoutes` is the same hand-maintained shape but smaller, stable, and not the reported bug (an agent does not parse the SPA shell as JSON). The same `RecordingRouter` primitive can derive it in a follow-up.

## Testing

- `TestMountAuthed_…`: a route registered through the authed path is reachable as the session auth failure (JSON), not the SPA HTML catch-all; a never-registered route still falls through to the SPA.
- `RecordingRouter` unit tests (records + forwards; `Patterns()` is a copy; `*http.ServeMux` satisfies `Router`).
- Gates green locally: `go build ./...`, `go test` across httpserver + cmd + all five contexts, golangci-lint (custom) 0 issues, `spectrace check --strict` 423/423, `openspec validate --strict`, dash lint.

openspec change `derive-authed-route-allowlist` (delta to `server-rest-api`) documents the invariant; the test carries its spectrace marker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically derive session-protected route allowlist from registered routes, eliminating manual maintenance and the risk of routes being registered but not properly protected.
  * Ensure unauthenticated requests to protected API routes return authentication-failure JSON responses instead of potentially falling through to the UI.

* **Refactor**
  * Updated route registration APIs across multiple services for consistency and reliability.

### DIFF
--- a/openspec/changes/derive-authed-route-allowlist/.openspec.yaml
+++ b/openspec/changes/derive-authed-route-allowlist/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-21
--- a/openspec/changes/derive-authed-route-allowlist/design.md
+++ b/openspec/changes/derive-authed-route-allowlist/design.md
@@ -1,0 +1,26 @@
+# Design: derive the authed-route allowlist (#463)
+
+## The constraint that shapes the fix
+
+Go's `http.ServeMux` exposes no API to enumerate the patterns registered on it (the pattern map is unexported, and Go 1.22's method+path routing did not add introspection). So neither "derive the allowlist" nor "test that every registered route is reachable" can inspect the API mux after the fact. Both have to capture patterns *at registration time*. The only way to capture them is to have `RegisterAuthedRoutes` register onto something we control instead of a concrete `*http.ServeMux`.
+
+## Decision: a Router interface + a recording wrapper, then derive
+
+- `httpserver.Router` is the minimal subset the registration bodies use: `Handle` + `HandleFunc`. `*http.ServeMux` already satisfies it, so changing the parameter type from `*http.ServeMux` to `httpserver.Router` leaves every call site compiling and every method body unchanged. The change is wide (every authed registration method, and the operator-handler `RegisterRoutes` the context bootstraps delegate to) but shallow and mechanical.
+- `httpserver.RecordingRouter` wraps a `Router`, forwards every `Handle`/`HandleFunc` to it, and records the pattern. `buildMux` registers all contexts through one `RecordingRouter` over the API mux, then mounts exactly `Patterns()` on the session-protected boundary.
+
+Because the mounted set *is* the registered set, a route can no longer be registered-but-not-mounted. The bug class is eliminated, not merely guarded, which is why this is preferred over a test that only asserts the hand-maintained list stays in sync.
+
+## Alternatives considered
+
+- **A test that re-lists the routes.** Moves the drift into the test (the test's route list can fall out of sync just like the slice did). Rejected: it does not make drift impossible.
+- **A data-driven route table per context** (`var routes = []Route{...}` consumed by both registration and an exporter). More invasive than the interface seam (rewrites how every handler registers) for no extra safety over recording.
+- **Reflecting/AST-parsing the source for patterns.** Fragile and indirect. Rejected.
+
+## Testability
+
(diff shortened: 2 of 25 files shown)


getvictor:
@coderabbitai review

Issue #450, opened 2026-06-19
QA: v0.3.0 RC validation for agent + extension changes (live edr-qa)


getvictor:
## Why

Per `CLAUDE.md`, any agent/extension change touching ESF, XPC, or the event wire format MUST be exercised on a live macOS VM (the system/VM layer) before RC. The v0.3.0 line landed several such changes whose per-PR unit/integration tests do not cover them end-to-end on a real enrolled host. This is the RC-gate checklist: run on **edr-qa** (SIP on + Gatekeeper on, signed/notarized pkg) before cutting the v0.3.0 RC.

History note (from #399): the v0.2.0-rc.1 edr-qa validation checked `systemextensionsctl` display names + ES health post-upgrade but did NOT verify the NE reconnected to the agent, so edr-qa was likely silently in the stale-NE state until its next reboot. The NE-after-upgrade scenario below is added to the standing upgrade checklist so this is not missed again.

## Setup

- edr-qa, SIP on + Gatekeeper on. Install via the **signed + notarized v0.3.0 RC pkg** (NOT a binary swap).
- For the upgrade scenario, also have the prior shipped pkg (v0.2.x) to upgrade FROM.
- Baseline: agent enrolled, both extensions activated, exec/fork/network/dns telemetry flowing to the server.

### 1. #399 - NE upgrade reboot signal (agent receiver + host app)
- [ ] Install vN; confirm `network_connect` + `dns_query` telemetry is flowing.
- [ ] Upgrade to the RC via the signed pkg; confirm `systemextensionsctl list` shows the OLD NE version `terminated waiting to uninstall on reboot`.
- [ ] Confirm the host app emits the reboot-required log line ("Fleet EDR upgrade staged; reboot to finish the extension cutover ...").
- [ ] Confirm the agent emits the distinct WARN (`hint=reboot_required_after_upgrade`, "network extension XPC registration is stale after an upgrade; reboot ...") instead of the bare `xpc_bridge_connect failed`, and does NOT spam it (once per stale episode).
- [ ] Confirm a FRESH install (no prior version) does NOT emit the reboot signal (no spurious prompt).
- [ ] Reboot; confirm the NE reconnects to the agent and network/dns telemetry resumes.

### 2. #398 - uploader keeps telemetry on a blanket edge rejection
[...]

PR #486, opened 2026-06-22
feat: kernel-cached app-control ALLOW (#209) + gzip telemetry upload (#405)

### LINKED ISSUE #209: App control: pass cache=true on AUTH_EXEC ALLOW for stable identities (kernel result cache disabled today)
## Summary

All three `es_respond_auth_result` calls in `extension/edr/extension/ESFSubscriber.swift` pass `false` for the cache
parameter:

- Line 119: team_id failsafe ALLOW
- Line 132: snapshot-matched DENY
- Line 153: post-decision default ALLOW (including the cold-cache miss)

The DENY at line 132 with `cache: false` is correct: if the operator removes a block rule, we want subsequent execs to
re-consult the callback instead of being permanently denied by the kernel cache.

The two ALLOW paths with `cache: false` are not. Every fork-exec storm of the same binary (Xcode build, npm install,
make, ASN.1 fuzzing, anything that re-execs a compiler / linker / interpreter in a tight loop) re-enters the handler.
Under load this is a real per-event cost (lock acquisition, snapshot read, optional FileHashCache lookup, libdispatch
hop for the lazy fill) that the kernel cache exists specifically to elide.

## What Santa does

Santa passes `cacheable = YES` for ALLOW on stable identities (binaries with a CDHASH or BINARY identity that is not
expected to change). The kernel then short-circuits subsequent exec auth queries for the same identity until the
binary's on-disk state mutates. Santa's cache invalidation is driven by ES's own notify events (write, unlink,
rename), so the kernel cache stays correct without the userspace handler being in the hot path.

## Proposed change

In `handleAuthExec`:

1. Failsafe ALLOW (currently line 119): pass `cache: true`. Same identity (`teamID == extensionTeamID`) produces the
   same answer for the lifetime of the binary on disk.
2. Post-snapshot ALLOW (currently line 153): pass `cache: true` if a hash was successfully looked up in the cache
   and no rule matched. The identity is stable and the rule snapshot is the same input. If the snapshot changes,
   invalidate via the existing snapshot-load path (`ApplicationControlStore.didLoadSnapshot` can call into a thin
   wrapper around `es_clear_cache(client)`).
3. Cold-cache ALLOW (the bypass branch in the separate cold-cache issue): keep `cache: false`. We do not yet know the
   identity. Caching this ALLOW would mean the kernel never asks again, defeating the point of the deferred fill.
4. DENY (line 132): keep `cache: false`. Rule removal must take effect on the next exec.

The cache invalidation hook on snapshot reload is the load-bearing piece. Without it, a rule added after a binary has
been ALLOW-cached at the kernel will never fire. Wire it via `es_clear_cache(client)` whenever the snapshot's
`policyVersion` advances.

## Scope and references

- `extension/edr/extension/ESFSubscriber.swift:108-154` (handler)
- `extension/edr/extension/ApplicationControlStore.swift` (snapshot loader, where invalidation hook lives)
- ES doc: `es_clear_cache` is the supported way to flush the kernel auth cache on snapshot reload.

## Milestone rationale

v0.2.0. The current posture is correct, just slow under fork-exec storms. The fix touches the snapshot-reload path
(invalidation) and needs a regression test that asserts `es_clear_cache` is called when the policy version advances.
Not pilot-blocking, but the first thing to land when we start measuring CPU per agent under load.

### LINKED ISSUE #405: Optimize agent->server telemetry upload: compression, persistent connections, modern transport

getvictor:
## Summary

Two independent efficiency wins that share one macOS VM QA session, shipped together (closes #209, closes #405).

### #209 Kernel-cached ALLOW with snapshot-driven invalidation (extension, Swift)

The AUTH_EXEC handler answered every ALLOW with `cache: false`, so a fork-exec storm of an already-allowed binary (a compiler/linker/interpreter re-execed in a tight loop) re-entered the handler every time: lock acquisition, snapshot read, optional hash lookup the kernel cache exists to elide. Santa caches ALLOW for stable identities; we did not.

- A fully decided `.allow` and the self-allow failsafe now respond `cache: true`; undecided ALLOWs (cold-cache / deadline / read-failure) and every DENY stay `cache: false`.
- The cache flag is a pure, ES-free `authResultIsCacheable(AuthDecision)` so the no-EndpointSecurity test target can pin it.
- Load-bearing correctness: `ApplicationControlStore` flushes the kernel AUTH cache (`es_clear_cache`) on **every accepted snapshot swap** via the new `onSnapshotApplied` hook, so a cached ALLOW cannot outlive a rule change. Fired on acceptance (version advance, epoch re-sync, or policy retarget), broader than the issue's "version advance" because all three mutate the active ruleset.

### #405 gzip request compression for `POST /api/events` (agent + server, Go)

Compression slice of #405. Persistent-connection / transport modernization stay in #405.

- Agent gzips each batch and sets `Content-Encoding: gzip`. Event JSON is small and repetitive, so this cuts upload bandwidth several-fold on the insert-heavy ingest path. The raw JSON still drives the agent's 413 split-and-retry (which bisects by event count), so that recovery is unaffected.
- Server decompresses when `Content-Encoding: gzip` is present and caps the **decompressed** stream at `MaxIngestBodyBytes`, so a small compressed body cannot expand past the cap (decompression bomb → 413 `body_too_large`). A malformed gzip stream is rejected as 400 `invalid_gzip`, distinct from the oversize signal.
- The plaintext path is preserved (the demo-seed tool and any non-gzip caller keep working), so no agent/server version lockstep is required.

## Spec

Three OpenSpec deltas under `openspec/changes/exec-auth-cache-and-ingest-gzip/`: `extension-application-control` (ADDED), `agent-event-uploader` (MODIFIED), `server-event-ingestion` (MODIFIED). New SHALL/MUST scenarios carry test markers (spectrace 100%).

## Testing

- `go test` (uploader wire-shape + intake bomb/corrupt rejections); the gzip batch persists end-to-end against real MySQL (integration).
- `swift test` (cacheability mapping + store flush-on-accept / no-flush-on-reject); extension target compiles via xcodebuild (the `es_clear_cache` wire path is real-typed).
- `task lint:go`, `lint:go:newcode`, `lint:dashes`, `swiftlint`, `openspec validate --all --strict`, `spectrace check` all green.

### Live VM QA (edr-dev, 2026-06-22)

**#405 passed end-to-end.** A real agent (fresh seeded queue, bypassing the broken edr-dev XPC) gzip-uploaded to a from-branch server; events persisted in MySQL. curl A/B on the wire:

| Case | Result |
|---|---|
| gzip + `Content-Encoding: gzip` | 200 `{"accepted":1}` |
| same gzip bytes, no header | 400 `invalid_json` (proves the body is genuinely compressed) |
| 11 MB → 10,720-byte bomb + header | 413 `body_too_large` |
| corrupt gzip + header | 400 `invalid_gzip` |

**#209 ESF cache QA is deferred to a live ES client** (edr-dev's agent↔extension XPC is broken). Tracked as section 7 of the v0.3.0 edr-qa QA issue (#450): reduced AUTH_EXEC handler entries under a fork-exec storm, self-allow/platform binaries still run, and the `es_clear_cache`-on-policy-push invalidation.

## Summary by CodeRabbit

* **New Features**
  * Event telemetry uploads now use gzip compression for reduced bandwidth consumption
  * Fully decided allow authorization decisions are now cached in the kernel to improve performance
  * Kernel authorization cache is automatically flushed whenever application control policies are updated, ensuring stale cached decisions do not persist across ruleset changes

### DIFF
--- a/agent/uploader/uploader.go
+++ b/agent/uploader/uploader.go
@@ -4,6 +4,7 @@ package uploader
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
@@ -466,12 +467,37 @@ func (*requestEntityTooLargeError) Error() string {
 	return "server returned 413 (request entity too large)"
 }
 
+// gzipBytes returns body compressed with gzip at the default level. Default level (not BestCompression) is the deliberate
+// pick: the batch is tiny and CPU-bound work runs on the agent's drain tick, so the marginal ratio from a slower level is not
+// worth the cycles on the endpoint. A bytes.Buffer cannot fail to write, so the only error path is the writer Close flushing
+// the trailer, which is surfaced to the caller.
+func gzipBytes(body []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	zw := gzip.NewWriter(&buf)
+	if _, err := zw.Write(body); err != nil {
+		return nil, fmt.Errorf("gzip write: %w", err)
+	}
+	if err := zw.Close(); err != nil {
+		return nil, fmt.Errorf("gzip close: %w", err)
+	}
+	return buf.Bytes(), nil
+}
+
 func (u *Uploader) doUpload(ctx context.Context, url string, body []byte) error {
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
+	// gzip the JSON batch on the wire. Event payloads are small, repetitive JSON (shared envelope keys, enum-like fields), so
+	// compression cuts upload bandwidth several-fold on the insert-heavy ingest path (#405). The raw `body` still drives the
+	// caller's 413 split-and-retry (which bisects by event count), so that recovery path is unaffected by the wire encoding.
+	compressed, err := gzipBytes(body)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(compressed))
 	if err != nil {
 		return err
 	}
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Content-Encoding", "gzip")
 	if u.cfg.TokenFn != nil {
 		if tok := u.cfg.TokenFn(); tok != "" {
 			req.Header.Set("Authorization", "Bearer "+tok)
--- a/agent/uploader/uploader_test.go
+++ b/agent/uploader/uploader_test.go
@@ -2,6 +2,7 @@ package uploader
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -56,7 +57,7 @@ func TestUploadBatch(t *testing.T) {
 			return
 		}
 
-		body, _ := io.ReadAll(r.Body)
+		body := readUploadBody(t, r)
 		if err := json.Unmarshal(body, &received); err != nil {
 			t.Errorf("unmarshal: %v", err)
 		}
@@ -267,11 +268,7 @@ func TestUpload_BoundedBatchSize(t *testing.T) {
 	// against the main-goroutine read after drainOnce. int64 avoids the gosec int->int32 overflow lint.
 	var receivedCount atomic.Int64
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		body, err := io.ReadAll(r.Body)
-		if err != nil {
-			t.Errorf("read body: %v", err)
-			return
-		}
+		body := readUploadBody(t, r)
 		var events []json.RawMessage
 		if err := json.Unmarshal(body, &events); err != nil {
 			t.Errorf("unmarshal: %v", err)
@@ -316,11 +313,7 @@ func TestUpload_DrainUntilCaughtUp(t *testing.T) {
 	// atomic because the httptest handler runs on its own goroutine; int64 avoids the gosec int->int32 lint.
 	var requests, events atomic.Int64
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		body, err := io.ReadAll(r.Body)
-		if err != nil {
-			t.Errorf("read body: %v", err)
-			return
-		}
+		body := readUploadBody(t, r)
 		var batch []json.RawMessage
 		if err := json.Unmarshal(body, &batch); err != nil {
 			t.Errorf("unmarshal: %v", err)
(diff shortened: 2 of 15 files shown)


getvictor:
@coderabbitai review

getvictor:
@coderabbitai review

getvictor:
@coderabbitai review

PR #501, opened 2026-06-24
feat(observability): route-tier OTel head sampling with runtime controls (#374)

### LINKED ISSUE #374: Reduce OTel export volume: head sampling + Fleet-style dynamic/runtime controls
## Summary

OTel export volume is unbounded today: there is no explicit trace sampler, so the SDK default (parent-based, always-on) records and exports every span. Fine at pilot volume, unsustainable at fleet scale and already costly in prod. We need (a) head sampling to cap the baseline volume and (b) Fleet-style dynamic / runtime controls so operators can dial telemetry up or down without a redeploy.

## Why now

- The prod EDR server is currently emitting `traces export: exporter export timeout: ... 502 (Bad Gateway)` against the collector. High span volume against a constrained collector makes this worse.
- For an EDR, metrics (not sampled spans) must be the source of truth for p99 / alerting once sampling is on.

### 1. Head sampling
- Add an explicit sampler, `ParentBased(TraceIDRatio(ratio))`, with the ratio tunable via env (e.g. `EDR_OTEL_TRACE_SAMPLE_RATIO`, default keeps a small percentage of healthy traces).
- Consider tail sampling in the collector as a follow-up (keep all error/slow traces + a small % of healthy ones).
- Document that p99/alerting reads from metrics, not spans, once sampling is on.

### 2. Dynamic / runtime controls (Fleet-style)
Fleet exposes runtime knobs for observability rather than requiring a restart. Mirror that:
- Runtime-adjustable log level (raise to DEBUG for a window, drop back) without redeploy.
- Runtime-adjustable trace sample ratio.
- A clear precedence/order for env defaults vs runtime overrides.

(Fleet reference: `~/work/fleet` logging + tracing configuration for the exact knob shapes.)

## Relationship to #348

#348 section 3 ("Trace sampling strategy") already proposes head/tail sampling as one of three v0.2.0 observability workstreams. This issue is the focused, standalone version of that workstream **plus** the dynamic-controls ask, which #348 does not cover. Either fold this into #348 section 3 or track it as the implementation issue for that section, but the dynamic-controls piece is net-new.

## Acceptance

- Trace sampler configured, ratio tunable via env, documented, with the metrics-are-authoritative-under-sampling note.
- A runtime path to adjust log level and sample ratio without a redeploy, matching Fleet's approach.
- Spec delta under `openspec/changes/` for the observability-instrumentation capability.

getvictor:
## Why

OTel trace export is unbounded: no sampler is configured, so the SDK default (parent-based, always-on) records and exports every span. Production is already failing trace export with `502 (Bad Gateway)` against the collector, and there's no way to dial telemetry down (during an export incident) or up (to debug) without a redeploy. Closes #374.

## What

Route-aware **head sampling** with four tiers, runtime-adjustable without a redeploy:

| Tier | Routes | Default ratio |
| --- | --- | --- |
| HighVolume | high-frequency agent data plane (`POST /api/events`, agent `GET /api/commands` poll, `POST /api/token/refresh`) | 0.01 |
| Standard | operator/UI `GET` reads | 0.1 |
| Full | everything else: writes, admin mutations, unclassified, and low-frequency load-bearing agent routes like `POST /api/enroll` | 1.0 |
| Drop | liveness/health probes (`/livez`, `/readyz`, `/health`) | never exported (wins over force-full) |

Wrapped in `ParentBased` so a sampled parent forces its children sampled. Ratios + a `force_full` incident toggle live in a `trace_sampler_settings` singleton table (CHECK-bounded, seeded with the compile-time defaults), polled by every replica each 60s, and adjusted via `GET`/`PATCH /api/settings/tracing` (admin + super_admin, the new `tracing.manage` action).

## Architecture

- `internal/observability/tracing` — the sampler mechanism (Tier, Registry, RouteTierSampler, Settings, poller). Agent-safe; no server or context coupling.
- `server/tracingpolicy` — the EDR route-to-tier policy, shared by both binaries.
- **New `observability` bounded context** (a sixth context, ADR-0004 amendment) owns the settings table + admin API. It owns a schema and serves an authz-gated, audited route, which is the bounded-context shape; arch-go forbids context-free platform packages from importing any context, so the authz/audit dependency rules this surface out of the platform layer. It consumes `identity/api` for the chokepoint + audit recorder.
- `observability.Init` / `bootstrap.Init` thread an `Options.Sampler` through; both `fleet-edr-server` and `fleet-edr-ingest` build the sampler, register the policy, and start the poller.

## Notes for reviewers

- **Metrics stay authoritative under sampling**: p99/error-rate/request-rate read from the `http.server.request.duration` histogram and the `edr.*` counters (never sampled), not from spans. Documented in `docs/operations.md` and pinned by spec scenarios.
- Operator detail reads with a path param (e.g. `GET /api/alerts/{id}`) aren't classified (otelhttp emits the raw path) and fall to Full (100%) safe-by-default; the volume-driving routes are all parameter-free.
- The dynamic in-app knob is an interim; collector-side tail sampling remains a possible follow-up (noted in the issue).

## Spec

OpenSpec change `otel-route-tier-sampling` modifies the `observability-instrumentation` capability (route-tier sampling, runtime-adjustable settings, force-full, admin endpoint, probe-drop, metrics-authoritative). spectrace markers cover every new SHALL scenario.

## Summary by CodeRabbit

* **New Features**
  * Added configurable trace sampling with route-based tiers, including full, standard, high-volume, and drop behavior.
  * Introduced an admin settings page/API to view and update tracing ratios without redeploying.
  * Enabled sampling controls to refresh automatically at runtime across replicas.

* **Documentation**
  * Updated runbooks and architecture docs with trace sampling guidance, limits, and operational examples.

* **Bug Fixes**
  * Improved observability defaults and safeguards so invalid sampling values are rejected and probes remain unexported.

### DIFF
--- a/arch-go.yml
+++ b/arch-go.yml
@@ -91,6 +91,20 @@ dependenciesRules:
         - "**.coordination.leader"  # MySQL-advisory-lock leader coordinator: pipeline.Runner gates retention + process-TTL on it
         - "**.testdb"
 
+  # observability context internals (issue #374). A sixth bounded context (an amendment to ADR-0004's
+  # original five) that owns the runtime trace-sampler settings: its store + admin handler. It imports
+  # the shared sampler infra (internal/observability/tracing) for the Settings domain type, and
+  # identity.api for the authz chokepoint + audit recorder its admin endpoint funnels through.
+  - package: "**.server.observability.internal.**"
+    shouldOnlyDependsOn:
+      internal:
+        - "**.server.observability.internal.**"
+        - "**.observability.tracing"
+        - "**.identity.api"
+        - "**.attrkeys"
+        - "**.httpserver"
+        - "**.testdb"
+
   # ----- testkit packages -----------------------------------------------------
   #
   # testkit is a curated test-fixture surface for each bounded context.
@@ -161,6 +175,16 @@ dependenciesRules:
         - "**.attrkeys"
         - "**.httpserver"
         - "**.sqlhelpers"
+        - "**.testdb"
+
+  - package: "**.server.observability.testkit"
+    shouldOnlyDependsOn:
+      internal:
+        - "**.server.observability.bootstrap"
+        - "**.server.observability.internal.**"
+        - "**.observability.tracing"
+        - "**.attrkeys"
+        - "**.httpserver"
         - "**.testdb"
 
   # ----- api/ packages: pure-types, no cross-context internals ----------------
--- a/docs/adr/0004-modular-monolith-bounded-contexts.md
+++ b/docs/adr/0004-modular-monolith-bounded-contexts.md
@@ -29,6 +29,8 @@ Adopt a modular-monolith layout with **five bounded contexts** under `server/`:
 | `endpoint`  | Host enrollment + host-token verification. Tables: `enrollments`.               |
 | `identity`  | Operator users, sessions, login. Tables: `users`, `sessions`.                   |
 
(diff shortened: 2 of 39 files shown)


getvictor:
@coderabbitai full review

PR #508, opened 2026-06-25
fix(packaging): deactivate system extensions via the host app on uninstall

getvictor:
uninstall.sh ran `systemextensionsctl deactivate`, which is not a real subcommand (exit 64) and whose developer verbs are blocked under SIP, so it removed nothing while the swallowed error reported success. It also named only the security extension and deleted the host app right after, orphaning both extensions with no way to tear them down under SIP.

Deactivate both extensions the way activation already works: submit the host app's `deactivate` request in the console user's GUI session via launchctl asuser, before removing the app. Verify the result from live systemextensionsctl state and report it truthfully: clean removal, reboot to finish, or still active with cause-specific guidance. On an MDM-managed host macOS refuses a local deactivation (authorizationRequired); the script keeps the app and tells the operator to remove the system-extension profile via MDM.

Drop the same bogus systemextensionsctl call from preinstall: upgrades replace the extension in place via re-activation and never need a deactivation.

Verified on edr-qa (SIP on): unmanaged uninstall stages both extensions for removal on reboot; MDM-managed uninstall is refused and prints the profile guidance.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved uninstall behavior for macOS so system extensions are deactivated more reliably during removal.
  * Added better handling for managed Macs, including clearer guidance when MDM profiles prevent local extension removal.
  * Prevented uninstall hangs and made success detection more accurate.

* **Documentation**
  * Updated installation, upgrade, and uninstall docs with clearer steps, expected outcomes, and recovery guidance.


getvictor:
@coderabbitai full review

Issue #427, opened 2026-06-18
Migrate the event store to ClickHouse (primary telemetry store)


getvictor:
## Summary

Migrate the event store from MySQL to ClickHouse as the primary backing store for high-volume telemetry (`events`, and the per-process tables that hang off it). This is the migration, distinct from #408, which only reduces MySQL disk growth and *evaluates* ClickHouse. This issue covers committing to and executing the move once #408's evaluation lands.

## Why

- **Disk + ingest scale.** #408 documents `events` growing ~8-10 GB/day/host on the Render pilot; MySQL row/index overhead and B-tree write amplification do not scale to the 500-agent target (#203). ClickHouse's columnar storage + compression is the industry-standard answer for append-mostly security telemetry.
- **Hunting depends on it (sequencing).** The v0.5.0 hunting/investigation epic (#415) is query-heavy: process search with filters + cursor pagination + sort (#418), events-over-time histogram (#420), server-side sibling aggregation (#416), tree virtualization (#421). Building those against MySQL and then re-building on ClickHouse is wasted work. Storage moves in v0.4.0 so hunting in v0.5.0 is built once, on the right engine.

## Prerequisites

- [ ] #408 cheap wins shipped (drop `snapshot_heartbeat`, coalesce network/DNS, index/PK diet) and ClickHouse evaluation concluded with a go decision.
- [ ] #203 scale-testing harness available to validate ingest + query under a realistic 500-agent load profile.

## Scope (to refine after #408 evaluation)

- Dual-store boundary: which tables move to ClickHouse (high-volume append: `events`; likely the derived process/network/DNS views) vs. which stay in MySQL (relational/transactional: identity, role_bindings, policies, alerts state).
- Ingest path: writer changes, batching, async insert semantics, backpressure.
- Query layer: repository abstraction so the hunting endpoints (#418/#420/#416) target ClickHouse; keep cross-context API boundaries (ADR-0004) intact.
- Retention: replace the MySQL retention jobs (#360) with ClickHouse TTL.
- Migration + rollback story; deployment/topology impact (single-VM quickstart and Render).
- ADR for the dual-store decision.

## Out of scope

- The cheap MySQL reductions and the evaluation itself (those are #408).

Related: #408 (reduce + evaluate), #203 (scale testing), #415 (hunting epic that consumes this), #360 (MySQL retention job this replaces).

getvictor:
Done. The event store is ClickHouse as of the cutover:

[...]

Issue #203, opened 2026-05-20
Soak, chaos, and 500-agent scale testing


getvictor:
Tracking the testing work that the UAT plan in `ai/uat/plan.md` explicitly defers out of the current milestone list.
The first scale baseline (100 simulated agents, 30 minutes) ships inside the UAT plan as M12; this issue covers
everything beyond that.

## Scope

Once M1 to M12 of the UAT plan have landed and the 100-agent baseline is recorded under `ai/uat/baselines/`, the
following layers are still missing:

### Soak

- 10 to 100 simulated agents, 24 hours, single server + MySQL.
- Assertions: no RSS growth on server beyond baseline + 10%, no queue backlog on agents, no goroutine leaks
  (compare `pprof` snapshots at t=0, t=1h, t=24h), MySQL row growth linear with event volume.
- Cadence: weekly on `main`, plus once per RC.

### Larger scale

- Jump from M12's 100 simulated agents to 500.
- Re-baseline p50 / p95 / p99 ingest latency, server CPU and RSS, MySQL CPU and disk I/O.
- Target SLA TBD, but the working assumption is p99 ingest < 250ms at the HTTP boundary and < 5% CPU per 100 agents
  on the server box.
- Pilot is sized at 10 to 500 endpoints per `ai/mvp/plan.md`; the 500-agent number lets us prove the upper end of the
  pilot envelope before customer onboarding.

### Chaos

Scripted disruptions during a soak run, asserting recovery:

- MySQL restart mid-batch: agent queue must replay cleanly, no event loss.
- Server restart: same expectation, plus the UI must surface a degraded-state banner during the outage window.
- TLS cert rotation while agents are connected: agents pick up the new cert on next dial without manual intervention.
- OIDC IdP outage: break-glass remains usable; existing sessions survive the outage.
- Network partition between agent and server: agent backs off, queue grows, server side experiences the recovery
  fan-in as a load spike that must not OOM it.

### Tooling

- Reuse the fake-agent library and headless binary shipped in M1 to M4 of the UAT plan.
- New `tools/chaosctl/` (Go) to orchestrate the disruptions deterministically. Driven by a YAML scenario format
  similar to the L6 efficacy corpus but with time-coded fault injections instead of attack steps.

## Out of scope here

- Anything covered by L0 to L6 in `ai/uat/plan.md`. This issue is strictly the "FUTURE" tier.
- External red-team / pen test. Separate workstream.

## Definition of done

[...]

PR #532, opened 2026-06-27
feat(visibility): hard-switch the event store to ClickHouse (#427)

getvictor:
## Summary

The atomic hard switch of the event store from MySQL to ClickHouse (ADR-0015, OpenSpec change `clickhouse-event-store`). Events now live in a durable ClickHouse archive (history + per-process correlation) plus a small MySQL work queue (the processing backlog), instead of one MySQL `events` table. MySQL keeps the control plane (process graph, alerts, hosts, settings). No data migration: pre-upgrade event history is not carried over; alerts and their self-contained evidence are preserved.

**`EDR_CLICKHOUSE_DSN` is now required** — `fleet-edr-server` and `fleet-edr-ingest` refuse to start without it.

## What changed

- **Ingest fan-out**: intake writes each accepted event to the ClickHouse `EventArchive` then enqueues it on the MySQL `EventLog`, returning 200 only after both. Archive-first so the durable lake has every event before it is queued; both writes idempotent by `event_id`, so an agent retry is safe. `ingested_at_ns` is server-stamped here (agents cannot set it).
- **Processor** claims from the `EventLog` queue (`FOR UPDATE SKIP LOCKED`, unchanged ADR-0011 semantics), not the `events` table.
- **Reads**: per-process network/DNS correlation and self-contained alert evidence (`EventsByIDs`) delegate to the archive; callers (process-detail query, DNS-C2 rule, alert store) are unchanged.
- **Retention**: ClickHouse-native TTL for events; the MySQL events delete sweep is gone (process-record pruning unchanged). Backlog gauge reads the queue.
- **Removed**: the dead detection-store event methods; migration `00007` drops the `events` table and the `alert_events -> events` FK.
- **Wiring**: both binaries open ClickHouse and build the `visibility` context; `fleet-edr-migrate` now applies the visibility MySQL schema too.
- **Demo**: `docker-compose.demo.yml` gains a ClickHouse service + `EDR_CLICKHOUSE_DSN`; the seeder slides archived-event timestamps on restart via a ClickHouse mutation. One compose file serves both demo legs (the released `latest` image ignores the unused var).
- **Docs**: CHANGELOG 0.4.0 + `install-server.md` document the required DSN; `task dev:server` sets it.

## Tests

Full default + `-tags integration` rework onto the queue + archive, plus new coverage: intake fan-out error branches + the `ingested_at_ns` stamp, the real-ClickHouse `EventsByIDs`, two-replica concurrent ingestion. spectrace 548/548 (2 scenarios exempted via the delta's `## REMOVED` index requirement); `openspec validate --all --strict` clean.

## Deferred (acceptance track, openspec 6.x)

The #203 500-agent scale baseline, the single-VM/quickstart deployment compose for ClickHouse, and `async_insert` tuning + PBT (annotated in `tasks.md`). Full demo e2e is the pre-release QA step.

## Summary by CodeRabbit

* **New Features**
  * Alerts now include the triggering event details in the alert-detail response.
  * Event retention is now handled automatically with time-based expiry in the new event archive.

* **Breaking Changes**
  * Existing deployments must provide the new event archive connection setting before the server or ingest service can start.
  * Event storage has moved to a new archive-backed flow, with related read paths updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,17 @@ Notable changes to Fleet EDR, newest first. This project follows [Semantic Versi
 
 ## [0.4.0] (unreleased)
 
+### Upgrade notes (action required)
+
+- **The event store is now ClickHouse.** The server stores raw events in a ClickHouse archive (durable history + per-process correlation) and a small MySQL work queue (the processing backlog), instead of a single MySQL `events` table. `EDR_CLICKHOUSE_DSN` (or `EDR_CLICKHOUSE_DSN_FILE`) is now **required**: the server and the standalone ingest service refuse to start without it. Stand up a ClickHouse instance alongside MySQL and point both binaries at it. This is a hard cutover with no data migration: pre-upgrade event history is not carried over (alerts and their self-contained evidence are preserved). MySQL still holds the control plane (process graph, alerts, hosts, settings).
+
 ### Added
 
 - **Self-contained alert evidence.** An alert now captures the full envelopes of its triggering events at creation time, so the evidence stays readable even after the underlying events age out of retention. The alert-detail API response gains an `events` array alongside `event_ids` (see `docs/api/openapi.yaml`).
+
+### Changed
+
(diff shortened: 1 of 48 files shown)


getvictor:
@coderabbitai review

PR #533, opened 2026-06-27
feat(visibility): batch-prune acked rows from the event work queue (#427)

getvictor:
## Summary

The ClickHouse cutover (#532) wired `EventLog.Ack` to mark rows `processed = 1` but nothing removed them, so the MySQL `event_queue` grew unbounded (confirmed on the demo: 1557 rows after a full seed). `CountPending` stayed correct, so detection was unaffected, but the table would grow without bound at fleet scale. The `Ack` doc already promised "a separate retention sweep" — this adds it.

For a high-throughput EDR the delete should not sit on the per-replica hot processing path, so `Ack` stays a cheap index UPDATE and a **leader-gated sweep does batched DELETEs off that path** — the same per-batch discipline the process-retention runner uses. The claim's `WHERE processed = 0` index seek is unaffected by piled-up done rows.

## Changes

- **`EventLog.PruneProcessed(ctx, batchSize)`**: batched, deadlock-retried DELETE of `processed = 1` rows until none remain. Only acked rows are removed (Nack reverts `processed = 2`, so a selected row is never concurrently resurrected). The ClickHouse archive history is untouched.
- **`QueuePruneRunner`**: a leader-gated loop with its own advisory lock and a 1-minute default interval, independent of age-based retention (the queue must be swept even when retention is disabled, and on a shorter cadence since it grows at the ingest rate). Wired into the detection `pipeline.Runner`.
- **Metric** `edr.event_queue.rows_pruned` so operators can watch the sweep keep pace with ingest.
- **OpenSpec**: adds the "Acknowledged work is pruned from the queue" scenario to the `clickhouse-event-store` delta's Decoupled-processing-pipeline requirement.

## Tests

`EventLog.PruneProcessed` (only-acked survivorship + multi-batch loop, integration), `QueuePruneRunner.Run` (count, error propagation, metric, unit), the recorder counter, and the detection + visibility integration suites (pipeline wiring regression-free). spectrace 548/548; openspec validate clean.

## Summary by CodeRabbit

* **New Features**
  * Added periodic pruning of processed events from the event queue, helping keep the queue smaller over time.
  * Introduced a new metric to track how many queue rows are removed during pruning.
  * Queue maintenance now runs with a default cadence during startup.

* **Bug Fixes**
  * Event ingestion now reports success only after the event is safely stored and queued.
  * Queue pruning preserves unprocessed and in-flight events while removing only acknowledged ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Notable changes to Fleet EDR, newest first. This project follows [Semantic Versi
 
 ### Changed
 
-- **Events move to a ClickHouse archive + a MySQL work queue.** Ingestion durably writes each accepted event to the ClickHouse archive and enqueues it on the MySQL queue, returning success only after both; the detection processor claims from the queue. Per-process network/DNS correlation and alert evidence read the archive. Event retention is now ClickHouse-native time-based expiry rather than a periodic MySQL delete (process-record pruning is unchanged).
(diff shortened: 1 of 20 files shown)


getvictor:
@coderabbitai review

Issue #535, opened 2026-06-28
Detection-processor throughput: batch the per-event DB round-trips + add intra-replica concurrency


getvictor:
## Summary

The detection processor is the per-replica scale bottleneck. It is I/O-round-trip bound on per-event database writes, not compute bound, so a single replica falls behind ingest at a few hundred hosts. This issue captures the fixes, all of which land in the delivery-independent processing path and therefore survive the eventual move to a streaming substrate (see [ADR-0016](../blob/main/docs/adr/0016-event-delivery-substrate.md)).

## Evidence (#203 scale test + pprof, 2026-06-28)

Single replica, isolated stack (MySQL + ClickHouse), 80/20 quiet/active-attack mix:

- Ingest scaled to 500 hosts cleanly: client p99 136ms, 0 errors, server RSS flat ~89MB.
- The processor did NOT keep pace: `event_queue` pending backlog grew linearly (0 to ~37k) at 500 hosts on one replica. 200 hosts stayed bounded. Single-replica breakeven is roughly 300 hosts at this (deliberately heavy) mix.
- Three replicas behind a balancer kept the backlog bounded under ~500 (the processor is not leader-gated and scales out via SKIP LOCKED).

pprof on a saturated single replica:

- **25% CPU utilization** over a 30s sample (7.58s of 30s). The replica is not compute bound.
- ~74% of on-CPU time is in `syscall`/`runtime.kevent` (netpoll): the processor goroutine is blocked on DB I/O.
- Wall-time concentrates in `graph.(*Builder).ProcessBatch`, which loops per event and calls `GetProcessByPID` (SELECT) then `UpdateProcessExec`/insert (UPDATE/INSERT). `database/sql.(*DB).retry` + `withLock` + `ResetSession` in the profile show each call acquires a pooled connection per statement.
- A 500-event batch is ~1000 to 1500 sequential, single-goroutine DB round-trips. That is the ~2.5s/batch.

## Proposed work (ranked)

[...]

PR #536, opened 2026-06-28
scale (#203): server-backlog gate for the M12 harness + ADR-0016 event-delivery substrate

getvictor:
## What

Two related outputs from the #203 scale-testing work, post-ClickHouse-cutover:

1. **Opt-in server-backlog gate for the M12 harness** (`test/scale`). The ingest p99 gate only measures the HTTP boundary; it can't see the detection processor falling behind. A local 500-host run showed ingest staying green (p99 136ms, 0 errors) while the MySQL `event_queue` backlog grew linearly to ~37k on one replica (bounded under ~500 across three replicas). This adds `--backlog-dsn` to poll the not-yet-acked `event_queue` depth and `--pass-max-server-backlog=N` to gate on it. Off by default: the per-PR smoke and the default baseline run never open a DB connection. It is the server-side counterpart to the agent-side `--pass-max-queue-depth` gate.

2. **ADR-0016: event delivery, DB work queue now, streaming substrate at scale.** Records the decision to keep the `event_queue` + SKIP LOCKED claim (ADR-0011) for the pilot envelope and invest in detection-processor throughput rather than a streaming backbone now, adopting a single-binary substrate (Redpanda / NATS JetStream, not Kafka) only when the fleet outgrows DB-as-queue. Includes the lever-transfer analysis (the DB-I/O batching + concurrency work survives a streaming migration; the queue/claim/ack/prune machinery is superseded; the backlog gate maps to consumer lag).

## Why

A pprof pass on a saturated single replica showed the processor is I/O-round-trip bound (25% CPU; time in per-event `GetProcessByPID` + `UpdateProcessExec` in the graph builder), not compute bound. The throughput fixes are tracked in #535; this PR ships the measurement tooling (the gate) and the architecture decision (the ADR) that frame that work.

## Notes

- `depthPercentile` moved from the build-tagged `runner_headless.go` to the always-compiled `runner.go` so the untagged `backlog.go` can share it (builds verified under both tag sets).
- `aggregateServerBacklog` is pure over (samples, opts) and unit-tested; it only ever flips `Pass` to false, composing after the ingest gates.
- The poll-interval default lives in one place (`startBacklogSampler` applies it when the interval is non-positive); the flag passes 0 rather than restating it.

Related: #203 (scale testing), #535 (processor throughput), ADR-0016, ADR-0011 (claim semantics), #427 (ClickHouse cutover).

## Summary by CodeRabbit

* **New Features**
  * Added optional server backlog monitoring for long-form scale runs, including configurable sampling and a pass/fail threshold.
  * Scale reports now include backlog percentiles, maximum depth, and sample counts.

* **Documentation**
  * Added guidance on the event-delivery architecture decision and updated ADR index.
  * Expanded scale-test docs with backlog gating and measurement details.

### DIFF
--- a/docs/adr/0016-event-delivery-substrate.md
+++ b/docs/adr/0016-event-delivery-substrate.md
@@ -1,0 +1,57 @@
+# 0016. Event delivery: database-backed work queue now, streaming substrate at scale
+
+- Status: Accepted
+- Date: 2026-06-28
+- Deciders: getvictor
+
+## Context
+
+After the ClickHouse cutover (ADR-0015), ingestion fans out each batch to the ClickHouse event archive plus a MySQL `event_queue` work queue. The detection processor claims from that queue with `SELECT ... FOR UPDATE SKIP LOCKED` (ADR-0011), processes the batch (process-graph materialization plus rule evaluation), acks it (`processed = 1`), and a leader-gated sweep prunes acked rows. This is the "database as a work queue" pattern: no message broker, all delivery state lives in MySQL.
+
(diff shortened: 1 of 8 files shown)


getvictor:
@coderabbitai review

Issue #518, opened 2026-06-26
Service accounts cannot write detection-config (exclusions/rule-settings): actor required at store layer


getvictor:
## Summary

An admin-roled service account passes the authorization gate for detection-config write endpoints but is then rejected at the store layer with `actor is required`, so service-account automation cannot create detection exclusions or rule-settings. This is the same class of defect as the recently fixed SSO-update issue (branch `fix-service-account-sso-update`, commit `41f8415a`): a write path assumes a human user actor where a service account has none.

## Impact

- Service accounts can read detection config and resolve alerts, but cannot manage exclusions or rule-settings programmatically.
- An operator wiring detection-config automation through a service account hits a 400 with a misleading `invalid_input` envelope (the request is well-formed; the actor shape is the problem).

## Steps to reproduce

1. Mint a service-account access token:
   ```sh
   curl -s -X POST https://<server>/api/oauth/token -H 'Content-Type: application/json' \
     -d '{"grant_type":"client_credentials","client_id":"sa_...","client_secret":"edrsa_..."}'
   ```
   The service account is bound to the `admin` role.
2. Create an exclusion with that bearer token:
   ```sh
   curl -s -X POST https://<server>/api/v1/detection-config/exclusions \
     -H 'Authorization: Bearer <token>' -H 'Content-Type: application/json' \
     -d '{"rule_id":"suspicious_exec","match_type":"parent_path_glob","value":"*/claude/versions/*","host_group_id":0,"reason":"benign dev tooling"}'
   ```

### Actual

```
HTTP 400
{"error":"detection_config.invalid_input","message":"detectionconfig: invalid request: actor is required"}
```

The same failure is expected for `PUT /api/v1/detection-config/rule-settings`.

### Expected

A service account holding the detection-config write permission can create exclusions and rule-settings. The audit row / `created_by` should attribute to the service account (for example `sa:<client_id>`) rather than requiring a numeric user ID.

## Root cause

The authz gate (`ActionDetectionConfigWrite`) admits the service account, but the persistence layer requires a non-empty actor identifier:

- `server/rules/internal/detectionconfig/service.go:146-151` `actorIdentifier()` returns `""` unless `actor.UserID > 0`. A service account authenticates by bearer token (claims `sub: sa_...`, role `admin`) and carries no `UserID`, so the identifier is empty.
- `server/rules/internal/detectionconfig/store.go:145` rejects the empty actor with `actor is required` (same gate for rule-settings at `store.go:196`).

[...]

PR #538, opened 2026-06-28
feat(identity): unified principal model + service-account attribution (#514, #518)

### LINKED ISSUE #514: Service-account actions are unattributable: give them first-class audit identity like users
## Summary

Service accounts (issue #376, ADR-0013) are second-class for auditability. When a service account performs a privileged mutation, we cannot record *which* service account did it: the audit row and the DB attribution columns both lose the identity. Users get full attribution; service accounts do not. We want service-account actions to be auditable to the same standard as user actions, following industry best practices for non-human principals.

## Where it breaks

1. **The authenticated actor carries no service-account identity.** `api.Actor` (`server/identity/api/authz.go`) has only `UserID int64`. `serviceaccounts.Authenticator.Authenticate` resolves a valid token to an `Actor` with `AuthMethod="service_account"` but `UserID=0` and **no client_id / subject**. After authentication, which service account is acting is simply gone from the request context.

2. **Audit rows for service-account actions are anonymous.** Audit attribution is keyed on the user: `AuditEvent.UserID *int64` + `ActorEmail string` (`server/identity/api/audit.go`). For a service account these are `0` / `""`. Example from a real SSO update by a service account: `audit recorded action=sso.config.updated actor_email="" edr.user.id=0`. The row exists but names nobody. (`ssoadmin.Handler.recordUpdate` hard-codes `UserID: &uid` from the actor's user id.)

3. **DB attribution columns can only reference users.** `oidc_config.updated_by` and `app_config.updated_by` are `BIGINT NULL` with FKs to `users(id)`. A service-account write has no user row, so it either violates the FK (the 500 we just hit) or, with the interim fix, records `NULL` (no operator). Either way the acting service account is not captured.

## Interim state (to be superseded by this work)

A stopgap fix makes service-account SSO updates record `updated_by = NULL` instead of `0` so the write no longer 500s (`ssoadmin/handler.go`, `bootstrap.go` apply closure threading a nullable `*int64`). That unblocks configuration but deliberately drops attribution. This issue is the real fix; the NULL behavior should be replaced once a principal model lands.

## Desired outcome

Every privileged action a service account takes is attributable to that specific service account, visible in the audit log/UI and in per-row attribution, to the same standard as a user action.

## Direction (industry best practices, to be finalized in design)

- **Unify the principal.** Model the acting principal as `(actor_type, actor_id)` where `actor_type in {user, service_account}` (a polymorphic actor reference, or a shared `principals` table both `users` and `service_accounts` key into). Carry it on `api.Actor` (add the service-account client_id/subject + a principal type) so it survives past authentication.
- **Audit always names a principal.** `AuditEvent` should record a stable principal id + type + display label and never collapse to empty/0. Service-account token issuance already audits the client_id as `TargetID` (`saadmin.recordIssuance`); extend the same identity to every SA-initiated mutation as the *actor*.
- **Attribution columns reference the principal, not just users.** For `updated_by`-style columns, pick one consistent model: reference a `principals` table, store the `(actor_type, actor_id)` pair, or treat the immutable audit log as the system of record for who-did-what and keep the FK column nullable. Apply it everywhere a mutation is attributed, not just SSO config.

## Acceptance criteria

- A service-account-initiated mutation produces an audit row that identifies the specific service account (client_id and/or display name), not `""`/`0`.
- Service-account attribution is queryable through the audit API/UI alongside user attribution.
- Per-row attribution (e.g. `oidc_config.updated_by` or its replacement) records the acting service account.
- The interim `NULL`-for-service-accounts behavior is removed or subsumed.

## Notes

This is a cross-cutting identity change (touches `api.Actor`, the audit pipeline, and attribution columns across contexts), so it ships with an `openspec/changes/<name>/` delta. Scope the principal model before implementing.

getvictor:
## What & why

Implements ADR-0017 (`openspec/changes/unified-principal-model/`): every authenticated actor resolves to a typed **principal** (user, service account, or system), so service-account actions are attributable to the same standard as users. Closes the two filed bugs:

- **#518** — service-account detection-config / app-control writes were rejected at the store layer with `actor is required`. Fixed: the store gates now accept the acting principal id (`svc_<id>`).
- **#514** — service-account mutations wrote audit rows that named nobody, and the `BIGINT`-FK attribution columns could only reference `users(id)`. Fixed: audit + per-row attribution now reference the principal id.

## What's in this PR (6 commits, each green)

- `PrincipalRef` foundation type (`usr_`/`svc_`/`sys` ids, `UserID()` accessor) — PBT-tested.
- `Actor.Principal` populated for both users (from the user row) and service accounts (from token claims, no DB read; `satoken` carries the principal id + name, stamped at the token endpoint).
- **#518 fixed:** detection-config + app-control attribution and store gates use the principal id.
- **principals spine + audit cutover** (migration 00005): `principals` table + backfill; `audit_events` rewritten to `(actor_type, actor_principal_id, actor_label snapshot)`, dropping the `users` join, so a deleted user's history stays attributable.
- **#514 attribution columns** (migration 00006): `oidc_config`/`app_config`/`service_accounts.created_by` → principal FK; #515 SSO `NULL` stopgap removed.
- observability `trace_sampler_settings.updated_by` → principal (migration 00002).

All migrations validated against the test DB; `go build ./...` and the touched server test suites are green.

## In-progress (follow-up commits on this branch before merge)

This is a staged cutover. To keep every commit green, two transitional fields are **intentionally** still present and will be removed in the contract commit before merge:

- `Actor.UserID` and `AuditEvent.UserID`/`ActorEmail` remain, bridged to the principal by the audit store (`actorOf`), so emitters not yet converted keep working. **This is local commit sequencing, not a shipped expand/contract** (ADR-0017 is a hard cutover).

Remaining before this is merge-ready:
1. Detection `alerts.updated_by` → principal (removes the `UserExists` FK-replacement + `ErrInvalidUserUpdater`).
2. Endpoint + response audit emitters → set `Actor`.
3. Convert identity's remaining `AuditEvent` emitters (oidc, breakglass, login, useradmin, saadmin, authz/engine) to `Actor`.
4. Contract commit: remove `Actor.UserID` + `AuditEvent.UserID`/`ActorEmail` + the bridge.
5. Principal-label resolver replacing the email closures.
6. #518 route-sweep regression + SA-attribution integration tests; openspec `tasks.md`.

Reviewers: please don't flag the `Actor.UserID` / `AuditEvent.UserID` duality as expand/contract; it is the transitional bridge called out above, removed in step 4.

## Checklist

- [x] **OpenSpec updated** — implements the merged `unified-principal-model` change (ADR-0017); deltas already on `main`.
- [ ] Tests added/updated — per-context tests updated to the principal model; the #518 route-sweep + SA-attribution integration tests land with the contract commit (see In-progress).
- [ ] Agent/extension ESF/XPC/wire change — N/A (server identity only).
- [x] **Docs updated** — ADR-0017 + the openspec bundle (on `main`).

## VM / RC notes (if applicable)

None. Server-side only; no agent or extension change.

Closes #514, #518.

## Summary by CodeRabbit

* **New Features**
  * Audit records now preserve a typed actor identity, improving attribution for users, service accounts, and system actions.
  * User, service account, and system activity now uses consistent principal IDs across the product.

* **Bug Fixes**
  * Audit history now remains readable even after the original user or service account is removed.
  * SSO, tracing, and app configuration updates now correctly show who made the change, including automated/system updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Notable changes to Fleet EDR, newest first. This project follows [Semantic Versi
 
 ### Changed
 
+- **Service-account actions are now fully attributable (unified principal model).** Every operator action is recorded against a typed principal (a user or a service account), so a service-account-initiated change names the specific service account rather than appearing as an anonymous or failed write. Service accounts can now manage detection-config exclusions and rule-settings (previously rejected with `actor is required`), and SSO / app-config / app-control / trace-sampler mutations they make are attributed to them. The audit-log API response identifies the acting principal for every row. See ADR-0017.
 - **Events move to a ClickHouse archive + a MySQL work queue.** Ingestion durably writes each accepted event to the ClickHouse archive and enqueues it on the MySQL queue, returning success only after both; the detection processor claims from the queue. Per-process network/DNS correlation and alert evidence read the archive. Event retention is now ClickHouse-native time-based expiry rather than a periodic MySQL delete (process-record pruning is unchanged). A leader-gated sweep prunes acknowledged events from the work queue in batches so it keeps to its in-flight working set; the `edr.event_queue.rows_pruned` metric reports its throughput.
 
 ## [0.3.0] (2026-06-26)
--- a/server/identity/api/audit.go
+++ b/server/identity/api/audit.go
@@ -91,6 +91,10 @@ const (
 // background ctx without the original span) still attributes the row
 // to the request that triggered it.
 type AuditEvent struct {
+	// Actor is the typed acting principal (user, service account, or system). It is the forward-looking attribution field: it names a
+	// principal for every actor kind, so a service-account action is auditable rather than collapsing to a zero user. See ADR-0017. During
+	// the principal-model cutover it rides alongside UserID/ActorEmail; the recorder snapshots its label onto the row.
+	Actor PrincipalRef
 	// UserID is the authenticated user, when one exists. Nil for pre-auth events (login_failed) so the audit row records "who attempted"
 	// without claiming "who was authenticated."
 	UserID *int64
--- a/server/identity/api/audit.go
+++ b/server/identity/api/audit.go
@@ -221,9 +221,15 @@ type AuditReader interface {
 // retrieval endpoint produces actionable rows in a single query, even after a user has been deleted from the users table (UserID then
 // resolves to a nil-or-empty email).
 type AuditRow struct {
-	ID         int64          `json:"id"`
-	OccurredAt time.Time      `json:"occurred_at"`
-	UserID     *int64         `json:"user_id,omitempty"`
+	ID         int64     `json:"id"`
+	OccurredAt time.Time `json:"occurred_at"`
+	// Actor is the typed acting principal recorded on the row (user, service account, or system). It is the principal-model attribution
+	// field; UserID/UserEmail below are derived from it for backward compatibility with existing API consumers. See ADR-0017.
+	Actor PrincipalRef `json:"actor"`
+	// UserID is the numeric users.id when the actor is a user, derived from Actor for back-compat. Nil for a service-account, system, or
+	// pre-auth-failure row.
+	UserID *int64 `json:"user_id,omitempty"`
+	// UserEmail is the actor's snapshot display label (a user's email, a service account's name) for back-compat display.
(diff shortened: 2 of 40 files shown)


getvictor:
@coderabbitai review

PR #539, opened 2026-06-28
refactor(identity): complete the principal cutover (shim removal) (#514)

getvictor:
## What & why

Completes the unified principal model (ADR-0017), following #538. #538 landed the principal spine + actor + audit attribution with a transitional bridge so each commit stayed green; this PR finishes the cutover and removes that bridge, so the model is **shim-free** and matches ADR-0017's hard-cutover decision.

## What changes

- **Detection `alerts.updated_by` → principal** (migration 00008): operator -> service -> store thread the acting principal id; the alert-status audit row sets `Actor`. The `UserExists` guard now runs only for a user principal; a service-account/system actor is trusted from the authenticated actor.
- **Contract step (shim removal):** drop `Actor.UserID`, `AuditEvent.UserID`/`ActorEmail`, and the audit-store `actorOf` bridge. Every remaining audit emitter (identity `oidc`/`breakglass`/`login`/`useradmin`/`saadmin`/`ssoadmin`/`jit`, the authz chokepoint, endpoint, response, and rules) now sets `Actor`. Pre-auth failures record a label-only principal (the attempted identifier). `AuditRow` keeps `UserID`/`UserEmail` as a derived back-compat projection for the retrieval API.
- **Rules attribution migration 00003:** rewrite existing `detection_*` + `app_control_*` attribution values (`user:<id>` -> `usr_<id>`, `system` -> `sys`) and change the column defaults from the free-form `'system'` to the `sys` principal id; the app-control Default-policy seed records the system principal.
- **#518 route-sweep regression test:** a service-account actor over the detection-config write routes asserts no `actor is required` rejection and `svc_<id>` attribution, guarding the surface against a future actor-shape regression.

## Deliberate refinements (vs the original openspec draft, documented in tasks.md)

- User/SA principals use a deterministic id (`usr_<users.id>` / `svc_<service_accounts.id>`) created in the row's transaction, rather than a stored `principal_id` FK column on `users`/`service_accounts` (avoids the insert-ordering chicken-and-egg; attribution columns still FK `principals(id)`).
- The detection `UserExists` FK-replacement check is retained for user principals only.

## Deferred (follow-up, not blocking)

A single principal-label resolver so a service-account author resolves to its name in the detection-config exclusions list. Attribution is already correct; only the SA display name shows the raw `svc_<id>` today. Recorded in `tasks.md`.

## Verification

`go build ./...`, the per-context integration suites (identity/rules/detection/observability/endpoint/response), `golangci-lint`, dash-lint, and markdownlint are green.

## Checklist

- [x] **OpenSpec updated** — completes the merged `unified-principal-model` change (ADR-0017); `tasks.md` updated.
- [x] Tests added/updated — #518 route-sweep regression, SA-attribution + snapshot tests, per-context test conversions.
- [ ] Agent/extension ESF/XPC/wire change — N/A (server-side only).
- [x] **Docs updated** — CHANGELOG entry landed with #538; ADR-0017 on main.

## VM / RC notes (if applicable)

None. Server-side only; no agent or extension change.

Relates to #514, #518 (both fixed by #538; this PR completes the cutover).

## Summary by CodeRabbit

* **New Features**
  * Alert updates and audit records now use unified principal IDs, so user, service account, and system actions are attributed consistently.
  * Detection and rules workflows now accept service-account and system principals in places that previously expected only numeric user IDs.

* **Bug Fixes**
  * Fixed alert status updates so the “updated by” value is preserved and stored in the new principal format.
  * Improved audit trail accuracy across login, OIDC, breakglass, user admin, and rule management flows.
  * Updated seeded/default attribution values to use the new system principal.


getvictor:
@coderabbitai review

Issue #176, opened 2026-05-16
Handle snapshot exec arriving after a live exit (graph builder)


getvictor:
## Problem

Discovered during PR #174 (issue #11) review by GitHub Copilot.

Race window: the extension's startup snapshot pass enumerates the live process table after `es_subscribe`. If a PID returned by the sysctl probe exits between the probe and the snapshot event's delivery to the server, the live `NOTIFY_EXIT` lands BEFORE the snapshot exec at the server. The server's `handleExit` calls `UpdateProcessExit` which is a no-op when no row exists yet. Then the snapshot exec arrives, hits `insertExecWithoutFork`, and creates a row with `exit_time_ns = NULL`. We now have a phantom "alive" row for a process that's actually dead, until the 6h TTL reconciler from issue #6 closes it.

Same root cause as issue #173 (snapshot rows getting TTL-reconciled), different angle: there #173 is about long-lived processes being closed prematurely; here it's about short-lived processes being kept alive falsely.

## Repro

Within the post-subscribe window (~1 second on the dev VM):

1. Sysctl probe sees PID P.
2. P exits. Live `NOTIFY_EXIT` event lands at server.
3. Server's `UpdateProcessExit` is a no-op (no row for P yet).
4. Snapshot enumeration emits an exec for P with `snapshot=true`.
5. Server creates a fresh row with no exit_time set.
6. UI shows P as alive until +6h.

## Proposed fix

Two layers, pick one or both:

**Server-side (preferred)**: When `handleExit` finds no row to update for a PID, buffer the exit (or a sentinel timestamp) for a short window (e.g. 30 seconds). When a subsequent snapshot=true exec arrives for the same PID and falls inside the window, set its `exit_time_ns` immediately and `exit_reason = host_reconciled`. Implementation can be a small in-memory map keyed by `(host_id, pid)` with the buffered exit timestamp.

**Extension-side**: Just-before-emit `kill(pid, 0)` check in `ProcessSnapshotEnumerator.makePayload`. If `ESRCH`, skip the emit. Still racy (the kill+emit window is also nonzero) but reduces the failure rate without server changes. Likely paired with the server-side fix as belt-and-suspenders.

## Out of scope

- The opposite race (live exec for a NEW process arrives during snapshot batch and gets clobbered) is already handled by the dedup added in PR #174 (`p.Snapshot && current != nil && current.ExecTimeNs != nil → drop snapshot exec`).
[...]

PR #541, opened 2026-06-28
perf(detection): batch processor DB round-trips + intra-replica concurrency (#535)

### LINKED ISSUE #535: Detection-processor throughput: batch the per-event DB round-trips + add intra-replica concurrency

getvictor:
## Summary

Fixes #535. The detection processor was the per-replica scale bottleneck: a single goroutine issuing ~1000-1500 sequential per-event DB round-trips per batch, leaving a saturated replica at ~25% CPU (I/O-round-trip bound, not compute bound). This PR implements the two code levers from the issue, both in the delivery-independent processing path that survives a future streaming cutover (ADR-0016).

### Lever 1: batch the per-event DB round-trips (`graph` + `mysql`)

`ProcessBatch` no longer loops one-or-more round-trips per event. It now:

1. **Bulk-loads** every candidate process row for the batch's `(host_id, pid)` set in one query (`mysql.LoadProcessesForKeys`).
2. **Folds in memory** against `graph.batchSession`, an overlay that reproduces the exact per-event SQL resolution semantics (timestamp ordering, PID reuse, exec-in-place, re-exec chain linkage, exec-without-fork, snapshot dedup, the #176 exit-before-snapshot buffer), including rows created earlier in the same batch.
3. **Flushes set-based** in one transaction (`mysql.FlushProcessBatch`): a multi-row INSERT for new rows + chunked `CASE`-keyed UPDATEs for modified rows.

The fork/exec/exit/heartbeat handlers were parameterized over a `processStore` interface, so the **identical handler code** drives both the in-memory session (production) and the store directly (the differential-test reference oracle). Poison isolation (#379) is preserved: malformed payloads drop during the fold; a permanent data error at flush degrades to a per-row apply that drops just the offending row; transient faults fail the whole batch atomically for retry.

### Lever 2: intra-replica concurrency (`pipeline` + `config` + `bootstrap`)

The processor runs `DefaultProcessConcurrency` (4) workers sharing one builder/engine, each claiming disjoint batches via the existing `SELECT ... FOR UPDATE SKIP LOCKED` claim (ADR-0011), with a full-batch drain loop. `OpenDB` now bounds the shared pool (`SetMaxOpenConns`/`SetMaxIdleConns`) so worker concurrency cannot exhaust MySQL. Both knobs are compiled constants, not env vars, per the minimal-configuration-surface requirement.

## Spec

`openspec/changes/detection-processor-throughput/`: a process-graph-builder equivalence requirement, a broadened availability SKIP-LOCKED requirement (now intra-replica) + a pool-ceiling requirement, and an extended configuration fixed-constants list.

## Testing

- **Differential PBT** (`processbatch_differential_test.go`): one big batch (exercises the in-memory overlay) == size-1 batches (true SQL semantics) over 8 crafted scenarios + a rapid generator. A genuine overlay-vs-SQL cross-check across the loaded/new-row boundary.
- **Poison + re-exec linkage** (`processbatch_flush_test.go`) and **8-worker drain** (`processor_concurrency_test.go`).
- Full `detection/`, `rules/`, `visibility/` integration suites pass; **efficacy corpus green** (no detection regression).
- `golangci-lint` custom + new-code strict gate: 0 issues; `openspec validate --all --strict`: 32/32; spectrace: 0 invalid refs.

## Out of scope (lever 3)

The #203 single-replica 500-host re-baseline (`--backlog-dsn` gate + pprof delta) is left for the scale lane: it needs the isolated MySQL+ClickHouse+fake-agent stack, not unit CI.

## Summary by CodeRabbit

* **New Features**
  * Added concurrent event processing within a single server instance, improving throughput.
  * Added batched process-graph handling to reduce database round-trips and better handle large event bursts.
  * Added safer fallback handling so malformed records can be skipped without stopping healthy work.

* **Bug Fixes**
  * Improved process row selection consistency when multiple records match.
  * Limited database connection usage to help prevent overload under higher concurrency.

* **Chores**
  * Tightened configuration behavior so operational settings are handled more consistently.


getvictor:
@coderabbitai review

Issue #459, opened 2026-06-20
Move detection allowlists (and other wrong-layer env-CSV configs) to a DB/policy-backed surface


getvictor:
## Context

Follow-up from the config-surface trim (`openspec/changes/trim-config-surface`, branch `config-trim`). That change removed dead/duplicate/internal-tuning knobs but deliberately **kept** the detection false-positive allowlists, because removing them would strip a pilot's only false-positive suppression with no replacement. They are kept, but they are in the **wrong layer** and should move.

## Problem

The four detection allowlists plus the disabled-rule list are configured as **boot-time environment-variable CSVs**:

- `EDR_LAUNCHAGENT_ALLOWLIST`
- `EDR_LAUNCHDAEMON_TEAMID_ALLOWLIST`
- `EDR_SUDOERS_WRITER_ALLOWLIST`
- `EDR_SUSPICIOUS_EXEC_PARENT_ALLOWLIST`
- `EDR_DISABLED_RULES`

This is the wrong shape for a stateless, multi-replica server (ADR-0010):

1. **Restart to change.** Tuning out a false positive during a pilot requires editing env and restarting every replica.
2. **No per-deployment audit trail.** Who added a path to the sudoers-writer allowlist, and when? Env vars carry no history; a DB/policy row does.
3. **Fleet-wide only.** There is no way to scope an allowlist entry to a host group, which is exactly what false-positive tuning often needs.
4. **Stateless-server invariant.** Durable cross-request operator state belongs in MySQL, not in per-replica boot env.

False-positive suppression is the #1 pilot pain for an EDR, so this is a real operator surface, not a nice-to-have.

## Proposal

Move these from boot env to a DB/policy-backed surface, editable through the admin UI / API and audited through the existing RBAC chokepoint, scoped to host groups where it makes sense. Mirror how Application Control policy already works (versioned, audited, host-group-assignable). Keep the env vars as a deprecated bootstrap fallback for one release, then remove.

## Also review (same smell)

While here, re-evaluate any other config that is durable operator state riding boot-time env and would be better in MySQL/policy. The recurring `docs/maintenance/tasks/config-surface-review.md` sweep (added in the same change) is the standing home for this check.

## Done when

- Allowlists + disabled-rule list are editable + audited via DB/policy (UI/API), not boot env.
- An OpenSpec change + delta documents the new surface.
- Docs (`docs/detection-rules.md`, `docs/operations.md`, `docs/install-server.md`) point at the new surface.

[...]

PR #550, opened 2026-06-29
fix(rules): one canonical name per detection across docs, catalog, and alerts (#519)

### LINKED ISSUE #519: Rule naming inconsistency: rule ID, doc title, and alert title diverge for the same detection
## Summary

Every detection rule exposes its name through three independent surfaces that do not agree, so a user sees a different string for the same detection depending on where they look:

1. **Rule ID** (machine identifier, snake_case) from `ID()`, surfaced in `/api/rules`, `alerts.rule_id`, exclusions, and the efficacy corpus.
2. **Doc title** from `Doc().Title`, surfaced in `/api/rules` and `docs/detection-rules.md`.
3. **Alert title** from the `Finding.Title`, surfaced in the alerts list and alert detail in the console.

For the LaunchDaemon-persistence rule the same detection is variously called:

- `privilege_launchd_plist_write` (ID)
- `LaunchDaemon persistence (BTM daemon registration)` (doc title)
- `LaunchDaemon persistence` (alert title)

A user triaging an alert, reading the docs, and writing an exclusion has to mentally map three strings to one detection. That is a usability and trust problem: the product should name a detection one way, consistently, the way SIEM and EDR catalogs do (a stable rule ID plus one canonical human-readable name reused everywhere).

## Scope: this is systemic, all rules are affected

| Rule ID | Doc title | Alert title(s) |
| --- | --- | --- |
| `privilege_launchd_plist_write` | LaunchDaemon persistence (BTM daemon registration) | LaunchDaemon persistence |
| `suspicious_exec` | Suspicious exec chain (non-shell → shell → temp/network) | Shell spawn with outbound network connection / Suspicious exec from temp path |
| `osascript_network_exec` | AppleScript dropper (osascript → curl/wget → temp exec) | osascript download-and-exec chain |
| `dns_c2_beacon` | DNS C2 beacon (suspicious process resolves a domain then connects to it) | DNS C2 beacon |
| `persistence_launchagent` | LaunchAgent persistence (launchctl load/bootstrap) | LaunchAgent persistence attempt |
| `sudoers_tamper` | Sudoers tamper (write to /etc/sudoers or /etc/sudoers.d/*) | Sudoers tamper |
| `credential_keychain_dump` | Keychain dump (security dump-keychain) | Keychain credential dump attempted |
| `dyld_insert` | DYLD injection on exec | DYLD injection env var set on exec |
| `shell_from_office` | Shell spawned by Microsoft Office | Shell spawned from Office app |
| `application_control_block` | Application control block | (computed per block) |

Two things stand out beyond the wording drift:

- **One rule, two alert titles.** `suspicious_exec` emits "Shell spawn with outbound network connection" on the network arm and "Suspicious exec from temp path" on the temp-exec arm. The same rule produces two different alert names, neither of which matches its doc title or ID.
- **The ID can be semantically stale.** `privilege_launchd_plist_write` no longer keys on a plist write at all; it keys on a BTM daemon registration (the doc says so). The ID itself misleads.

## Why it matters (industry best practice)

Detection catalogs (Sigma, Elastic, MITRE-aligned vendor rules) give each detection a stable identifier and one canonical human-readable name, and reuse that name across the rule catalog, the documentation, and the alert it raises. The alert a user sees should name the rule they can look up. Three different strings for one detection erodes that mapping and makes triage, tuning, and docs harder to trust.

## Proposed direction

- Define a single canonical display name per rule and reuse it for both `Doc().Title` and `Finding.Title`, so the docs, `/api/rules`, and the alert all show the same name. Where a rule has multiple arms (`suspicious_exec`), either use the one canonical name for both arms or document a deterministic "<rule name>: <arm>" derivation rather than two unrelated strings.
- Keep the snake_case `ID()` as the stable machine identifier. Renaming IDs is a breaking change: rule IDs are referenced by `alerts.rule_id`, detection-config exclusions, the efficacy corpus, openspec specs, and generated docs. Treat any ID rename (for example the stale `privilege_launchd_plist_write`) as a separate, migration-backed change, not part of the cosmetic-name alignment.
- Consider a small structural guard so the surfaces cannot drift again (for example a single `Name()`/`DisplayName()` on the rule that both the doc and the finding consume), and a catalog test asserting every rule's alert title is derived from its canonical name.

## Notes for the implementer

- This changes observable alert titles and the documented rule names, so it ships an `openspec/changes/<name>/` delta (proposal + tasks + spec delta) in the same PR.
- Audit all 10 rules under `server/rules/internal/catalog/`, not just the launchd rule.

getvictor:
## What & why

Closes #519. Every detection rule named itself three independent ways (`ID()`, `Doc().Title`, `Finding.Title`) that did not agree, so an operator saw a different string for the same detection depending on where they looked. `suspicious_exec` even emitted two unrelated alert titles, one per arm. Detection catalogs (Sigma, Elastic, MITRE-aligned vendor rules) instead give each detection a stable identifier plus one canonical human-readable name reused everywhere; the alert names the rule you can look up.

### Changes

- **`DisplayName()` on the `api.Rule` interface** is the single canonical name. Both `Doc().Title` and `Finding.Title` derive from it, so the docs, `/api/rules`, and the alert show one name mapped to one stable `ID()`.
- **Observable alert titles change** for the rules whose finding title diverged: `suspicious_exec` (both arms now "Suspicious exec chain"; which arm fired stays in the finding description, SIEM/Sigma convention), `persistence_launchagent`, `dyld_insert`, `shell_from_office`, `osascript_network_exec`, `credential_keychain_dump`. Documented titles lose the parenthetical implementation detail (it lives in `Summary`).
- **Rule IDs are unchanged.** Renaming an ID (e.g. the stale `privilege_launchd_plist_write`) is a separate migration-backed change: IDs key `alerts.rule_id`, exclusions, the efficacy corpus, and generated docs.
- **`application_control_block` is the one exemption:** it keeps its per-block computed title (`Application blocked: <binary>`) and `app_control:<n>` RuleID, since those alerts name the blocked binary and admin rule, not a catalog detection.
- **Guards against re-drift:** `TestAll_CanonicalDisplayName` asserts `Doc().Title == DisplayName()` and a clean (no-parenthetical) name for every rule; the fixture-replay harness and each positive-detection test assert `Finding.Title == DisplayName()`.
- Regenerated `docs/detection-rules.md`; fixed the OpenAPI example + a UI mock.

Ships the `canonical-rule-naming` openspec delta (the new "Canonical rule naming" requirement + scenario, referenced by a `spec:` marker in `registry_test.go`).

### Follow-up cleanups in this PR (second commit)

- `gen-rule-docs` `TestRenderConfigKnobsListed` asserted four `EDR_*_ALLOWLIST` env-var knobs no shipped rule declares anymore (tuning moved to the detection-config surface, #459) and failed on `main` independent of this branch; re-pointed at a synthetic rule so it pins the renderer, not the live catalog.
- Removed the DetectionConfig picker's now-dead parenthetical-stripping (`ruleLabel`): canonical titles never carry a parenthetical, so the picker shows the title verbatim.

## Checklist

- [x] **OpenSpec updated for behavior changes.** `openspec/changes/canonical-rule-naming/` delta added; `spec:` marker in `registry_test.go`.
- [x] Tests added/updated (catalog guard + fixture-replay + per-rule finding-title asserts + UI).
- [ ] Agent/extension change touching ESF, XPC, or the event wire format: N/A (server + UI only).
- [x] **Docs updated for user-facing changes.** `docs/detection-rules.md` regenerated; OpenAPI example updated.

## VM / RC notes

None. Server + UI only; no agent/extension/wire change.

https://claude.ai/code/session_01BxbWcMpjYFGpHZDU2d2mfm

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -1160,7 +1160,7 @@ components:
       properties:
         title:
           type: string
-          example: "Keychain dump (security dump-keychain)"
+          example: "Keychain credential dump"
         summary:
           type: string
           description: One-line elevator pitch shown in tooltips.
--- a/docs/detection-rules.md
+++ b/docs/detection-rules.md
@@ -15,20 +15,20 @@ Hand-edits to this file get overwritten on the next regeneration.
 
 | Rule ID | Title | Severity | ATT&CK |
 | --- | --- | --- | --- |
-| [`suspicious_exec`](#suspicious_exec) | Suspicious exec chain (non-shell → shell → temp/network) | high | T1059, T1105 |
-| [`persistence_launchagent`](#persistence_launchagent) | LaunchAgent persistence (launchctl load/bootstrap) | high | T1543.001 |
+| [`suspicious_exec`](#suspicious_exec) | Suspicious exec chain | high | T1059, T1105 |
+| [`persistence_launchagent`](#persistence_launchagent) | LaunchAgent persistence | high | T1543.001 |
 | [`dyld_insert`](#dyld_insert) | DYLD injection on exec | high | T1574.006 |
 | [`shell_from_office`](#shell_from_office) | Shell spawned by Microsoft Office | high | T1566.001, T1059.004 |
-| [`osascript_network_exec`](#osascript_network_exec) | AppleScript dropper (osascript → curl/wget → temp exec) | critical | T1059.002, T1105 |
-| [`credential_keychain_dump`](#credential_keychain_dump) | Keychain dump (security dump-keychain) | high | T1555.001 |
-| [`privilege_launchd_plist_write`](#privilege_launchd_plist_write) | LaunchDaemon persistence (BTM daemon registration) | high | T1543.004 |
-| [`sudoers_tamper`](#sudoers_tamper) | Sudoers tamper (write to /etc/sudoers or /etc/sudoers.d/*) | high | T1548.003 |
+| [`osascript_network_exec`](#osascript_network_exec) | AppleScript dropper | critical | T1059.002, T1105 |
+| [`credential_keychain_dump`](#credential_keychain_dump) | Keychain credential dump | high | T1555.001 |
+| [`privilege_launchd_plist_write`](#privilege_launchd_plist_write) | LaunchDaemon persistence | high | T1543.004 |
+| [`sudoers_tamper`](#sudoers_tamper) | Sudoers tamper | high | T1548.003 |
 | [`application_control_block`](#application_control_block) | Application control block | medium |  |
-| [`dns_c2_beacon`](#dns_c2_beacon) | DNS C2 beacon (suspicious process resolves a domain then connects to it) | high | T1071.004, T1568.002 |
+| [`dns_c2_beacon`](#dns_c2_beacon) | DNS C2 beacon | high | T1071.004, T1568.002 |
 
 ## suspicious_exec
 
-**Suspicious exec chain (non-shell → shell → temp/network)**  
+**Suspicious exec chain**  
 Flags a non-shell process that spawns a shell which, within 30 seconds, execs from /tmp or makes an outbound network connection.
 
 | | |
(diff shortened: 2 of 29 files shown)


getvictor:
@coderabbitai review

Issue #522, opened 2026-06-26
Observability: benign duplicate-key SQL writes (alert dedup, role-binding seed) surface as hasError trace spans


getvictor:
## Summary

Several SQL writes that are part of normal, expected operation surface as `hasError=true` trace spans (and driver-level errors), polluting trace-based error monitoring on `fleet-edr-server`. The clearest case is alert dedup: a re-fired finding deliberately collides with the unique key and is caught as the dedup path, but the failed `INSERT` still emits an error `sql.stmt.exec` span on every deduplicated alert. Because the server's whole observability story is OTel/SigNoz, this means you cannot build a trace `hasError` rate alert without it firing on routine activity.

## Observed

During a routine health check on the prod quickstart VM (6h window), the only error spans on `fleet-edr-server` were benign duplicate-key paths:

- `INSERT INTO alerts (...)` — one error span, on alert dedup.
- `INSERT INTO role_bindings (...)` — two error spans, at the two container start times in the window.

No functional impact (the app handles both), but each is a false positive for any error-rate signal built on traces, and they clutter the SigNoz error views.

## Root cause

`server/detection/internal/mysql/alerts.go:61-72` inserts the alert, then on duplicate-key error branches to `attachEventsToExistingAlert` as the dedup path:

```go
res, err := tx.ExecContext(ctx, `INSERT INTO alerts (...) VALUES (...)`, ...)
if err != nil {
    if isDuplicateKeyErr(err) {
        existingID, attachErr := s.attachEventsToExistingAlert(ctx, tx, a, eventIDs)
        return existingID, false, attachErr // dedup branch never creates a row
    }
    return 0, false, fmt.Errorf("insert alert: %w", err)
}
```

The duplicate is expected (it IS the dedup signal), but the driver returns an error before the app swallows it, so the instrumented span is flagged `hasError`. Every deduped alert emits one.

`server/identity/internal/rbac/role_bindings.go:75-78` (`BindRole`) inserts with no duplicate handling; re-binding an already-bound (user, role, scope) raises a duplicate-key error span. Observed at boot, consistent with an idempotent re-seed of an existing binding.

## Proposed fix

Replace insert-and-catch-the-duplicate with a race-safe idempotent upsert so the expected-duplicate case never raises a driver error:

[...]

PR #551, opened 2026-06-29
fix(observability): stop benign duplicate-key SQL writes raising hasE…

getvictor:
…rror trace spans (#522)

Several SQL writes that are part of normal operation deliberately collide with a unique key and swallow the resulting duplicate-key error as their control flow. The driver still raised the error first, so each routine collision surfaced as a hasError sql.stmt.exec span, making a trace error-rate alert on fleet-edr-server fire on routine activity.

Replace insert-and-catch-the-duplicate with race-safe idempotent upserts so the expected-duplicate case never reaches the error path (MySQL 8.4):

- InsertAlert: INSERT ... ON DUPLICATE KEY UPDATE id = LAST_INSERT_ID(id). LastInsertId() returns the existing row id on the dedup path; created is derived from RowsAffected (1 = inserted, 0 = matched existing, since the DSN does not set clientFoundRows). Drops attachEventsToExistingAlert, isDuplicateKeyErr, and the FOR UPDATE re-lookup: one round-trip, no span, identical dedup semantics across replicas (ADR-0010).
- rbac.BindRole: INSERT ... ON DUPLICATE KEY UPDATE user_id = user_id. The no-op update leaves the existing binding untouched, matching the prior swallow-the-duplicate semantics; unlike INSERT IGNORE a real FK/other error still surfaces. seed.bindSuperAdmin no longer branches on the dup.
- identities.UpsertWith (new) + breakglass.resolveLocalPasswordIdentityID: the redemption now upserts the local_password identity in one statement instead of insert-then-lookup-on-duplicate, so re-resolving a seeded identity no longer raises a span. InsertWith stays for JIT, where the duplicate-key error IS the concurrency-race signal ProvisionOrFind needs.

Audited the rest and left as-is, because the duplicate there is not benign routine traffic: the app-control create/rename and user-provision paths map the collision to a user-facing 409 conflict; the OIDC JIT identity-insert duplicate is a rare concurrency-race re-resolve (error+rollback is the mechanism); the detection retry classifier and the deadlock helper are not insert-and-catch (and processes has no unique key, so 1062 can't fire there).

Internal hygiene: dedup/seed/redemption behavior is unchanged, so no spec delta.

Claude-Session: https://claude.ai/code/session_01BxbWcMpjYFGpHZDU2d2mfm

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved idempotency for alert creation, identity setup, and role binding so repeated operations no longer create duplicates or fail on expected collisions.
  * Pre-existing identities are now correctly reused during breakglass setup, preserving the original identity record.
  * Reapplying the same role binding now becomes a no-op, leaving existing access details unchanged.

* **Tests**
  * Added integration coverage for pre-existing identity resolution and repeated role binding behavior.

### DIFF
--- a/server/detection/internal/mysql/alerts.go
+++ b/server/detection/internal/mysql/alerts.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/go-sql-driver/mysql"
 	"github.com/jmoiron/sqlx"
 
 	"github.com/fleetdm/edr/server/detection/api"
@@ -21,8 +20,9 @@ const alertEventsBatchSize = 500
 
 // InsertAlert creates an alert and links it to the given event IDs.
 // If a duplicate alert exists (same source, host_id, rule_id,
-// subject), the insert is skipped and the existing alert ID is
-// returned. Returns the alert ID and whether it was newly created.
+// subject), the existing row is matched and its ID is returned
+// without raising a driver error. Returns the alert ID and whether
+// it was newly created.
 //
 // Callers SHOULD set a.Source; a blank Source defaults to
 // AlertSourceDetection so existing catalog-rule call sites that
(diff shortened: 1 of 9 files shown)


getvictor:
@coderabbitai full review

PR #552, opened 2026-06-29
docs(response): propose push-based command channel + ADR-0010 gateway…

getvictor:
… carve-out (#477)

Replace the agent's GET /api/commands short-poll with a persistent push connection. This is the OpenSpec change + ADR amendment only; no code yet.

Scope (decided): build the persistent gRPC-bidi command-push stream with a thin gateway embedded in fleet-edr-server, routing delivery over the existing MySQL commands queue (in-process fast path on the connection-holding replica, 1s watch as the cross-replica floor). No NATS, no RTR, no mTLS: those stay later, evidence-gated steps of the scale-100k plan (Phase 3). The short-poll is retained as the degraded floor, so rollback is non-destructive and no schema changes.

- New capability agent-control-channel (8 requirements): connect-time host token auth, real-time delivery, outcome reporting + lifecycle, host-scoped isolation, connection-presence liveness, silent-failure reconnect, revocation teardown, idempotent at-least-once delivery.
- Modified agent-command-executor (stream preferred, poll is the floor), server-availability (stateless invariant carves out the gateway tier), server-configuration (1s command-watch interval is a fixed constant).
- ADR-0010 amended to name the control-connection gateway as the one sanctioned stateful tier (connections only, nothing durable, losable on reconnect without command loss).

Validated: openspec validate push-command-channel --strict; lint:dashes, lint:md, lint:md:prose clean.

## Checklist

- [ ] **OpenSpec updated for behavior changes.** If this PR changes observable behavior (a detection rule, the event
      wire schema, persistence semantics, an API/wire shape, or an emitted event), `openspec/specs/**` is updated (and
      an `openspec/changes/` proposal added), with a `spec:<id>` test marker for any new/renamed scenario. Only if this
      PR changes NO observable behavior (a comment / refactor / gofmt / dep bump that happens to touch a scoped path)
      may you assert `no-behavior-change` (label or `[no-behavior-change]` in the title) to clear the OpenSpec-sync gate.
      That is an assertion a reviewer will check, not a way to skip the spec for a real behavior change.
- [ ] Tests added/updated for the change (unit / integration / efficacy / UI as applicable).
- [ ] Agent/extension change touching ESF, XPC, or the event wire format: exercised on a live macOS VM before RC
      (flagged below).
- [ ] **Docs updated for user-facing changes.** If this PR changes a user-facing surface (the React UI, an HTTP
      handler, or a detection rule), `docs/` or `CHANGELOG.md` is updated in the same PR. Only if nothing a user sees
[...]

Issue #477, opened 2026-06-22
Push-based agent control channel: replace GET /api/commands polling (split from #405)


getvictor:
## Summary

Replace the agent commander's `GET /api/commands?host_id=...&status=pending` short-poll with a single persistent server-push channel so commands (and interactive RTR) reach the host in real time. Split out of #405 (telemetry-upload optimization), where this was tracked as "Option 4". It is a control-path concern, not telemetry-upload.

## Decision (single best-in-class approach)

One control channel: a **persistent gRPC bidirectional stream over server-TLS + cert pinning**, authenticated at connect with the **self-validating signed host token (#454)** carried in the connection metadata. We do NOT build a second transport.

- **Not mTLS.** Auth on the control channel is once-per-connection, not per-request, so the signed token over server-TLS + pinning is sufficient and avoids standing up a client-cert PKI. This matches the majors (CrowdStrike + Defender use token enrollment + server pinning; SentinelOne is the minority mTLS adopter). mTLS (ideally an agent-generated Secure Enclave key) stays an opt-in hardening upgrade on the same channel, not a prerequisite.
- **Not long-poll.** Long-poll is a fallback technique, not best-in-class; the majors all hold a persistent connection. Building it would mean a second protocol and permanent fallback maintenance on the control path.
- **Required for RTR.** Interactive live shell needs a persistent bidi stream regardless (request/response cannot do an interactive shell well), so the stream is built anyway.

## Current state

- `agent/commander/commander.go:105,131`: ticker-driven short-poll of `GET /api/commands`; result/ack via `PUT /api/commands/{id}` (`:294`).
- Latency floor is the poll interval; request volume scales with host count even when idle.

## Plan

1. **Interim: keep the existing short-poll.** It already works, is stateless, and gives a few-seconds latency on async commands (kill / isolate / rotate) that is fine at pilot scale. No new work; do not build long-poll.
2. **Build the persistent gRPC-bidi gateway** (the scale-to-100k control-channel design, Phase 3): a dedicated stateful gateway tier holding agent connections + a routing backplane (NATS, or over the event log) so any control-plane replica can deliver to the gateway holding host N's connection. Authenticate the stream at connect with the signed token. Move async commands onto the stream and retire the short-poll. Build interactive RTR on the same channel.
[...]

PR #553, opened 2026-06-30
feat(response): server side of push-based agent control channel (#477)

getvictor:
Implements the server half of the persistent gRPC control channel that replaces GET /api/commands polling (OpenSpec change push-command-channel, merged in #552). Server-first: this lands the gateway and serves it opt-in; the agent control client is a follow-up. Default-off, so nothing changes in the default deployment and the short-poll path remains the floor.

Foundation:
- internal/control: control.proto bidi-stream contract (Command down, Outcome up; oneofs for RTR-readiness) + generated Go; `task proto` regenerates it. grpc/protobuf promoted to direct deps; otelgrpc propagates trace context over stream metadata. PBT round-trip pins the wire shape.

Gateway (server/response/internal/gateway), the ADR-0010 sanctioned stateful tier:
- Connect-time auth interceptor reuses endpoint VerifyToken (ErrInvalidToken -> Unauthenticated, else Unavailable); pins host_id into the stream context.
- Registry holds at most one connection per host, evicting+closing a prior connection on reconnect.
- Watch loop (1s, host_id IN locally-held, index-scoped) is the cross-replica floor; in-process Notify fast path delivers a command queued on this replica immediately. Per-connection in-flight set dedups pushes.
- Outcomes flow through the unchanged UpdateStatus lifecycle; connection presence drives last-seen (liveness); a revoked/expired token tears the connection down within the revocation-snapshot bound.
- Bounded graceful Stop: long-lived streams are force-closed after a grace window so shutdown cannot hang.

Wiring:
- store.ListPendingForHosts (index-scoped) + service passthrough + fast-path notifier hook on Insert/InsertBatch.
- response.BuildControlGateway; served in main.go on a dedicated TLS listener when EDR_CONTROL_ADDR is set (reuses the server TLS 1.3 cert via the new httpserver.BuildTLSConfig). The HTTP serve path is untouched.
- arch-go allowlists internal/control for response.internal; dash-lint now skips machine-generated files.

Tests: bufconn delivery/outcome/auth/eviction/revocation, end-to-end TLS serve+stop, real-MySQL store/service queries, and the bootstrap wiring. go build, golangci-lint, dash-lint, and the affected unit + integration tests all pass.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **New Features**
  * Added a new agent control channel for persistent bidirectional communication with connected hosts.
  * Support for enabling the control gateway on a separate address, with optional TLS.
  * Improved TLS configuration handling, including certificate reload support.

* **Bug Fixes**
  * Control channel command delivery now uses a faster notification path and better handles reconnects and retries.
  * Pending commands are delivered in order and are tracked more reliably to avoid duplicate sends.

* **Tests**
  * Added coverage for protobuf round-trips, gateway behavior, TLS serving, and pending-command queries.


getvictor:
@coderabbitai full review

Issue #558, opened 2026-06-30
Share command idempotency state across poll and push transports (avoid kill_process re-execution)


getvictor:
## Summary

The agent runs one command executor per transport: the poll path (`commander`) and the push path (`controlclient`) each construct their own `commander.Executor`, and the idempotency `outcomeCache` is local to `controlclient`. Because the two paths don't share that cache, a command executed over the push channel can be re-executed by the poll path in a narrow window, which can corrupt the server-side command status for a non-idempotent command (`kill_process`).

Raised by Gemini in review of #556 (`agent/commander/executor.go`). Tracked as a follow-up rather than fixed in #556 because both fixes have scope/spec implications.

## The race

1. The control stream is up, so the commander suspends polling.
2. A `kill_process` command is pushed; `controlclient` acks, executes the kill, and starts to report `completed`.
3. The stream drops before the server processes the ack/outcome, so the command stays `pending` server-side.
4. The poll path resumes (degraded floor), fetches the still-pending command, and re-executes it with a fresh `Executor` that has no memory of the push-path `outcomeCache`.
5. Killing an already-dead PID returns `ESRCH`, which `executeKill` reports as `failed`, overwriting the intent of a kill that actually succeeded.

Mitigations already in place keep the blast radius small: polling is suspended while the stream is connected, and the server-side ack (sent before the side effect) largely closes the window. The remaining harm is a cosmetic status flip (`completed` -> `failed`) on an action that already did what was asked, not a safety issue (the process is dead either way). Still worth closing for status fidelity on operator actions.

## Options (decide during implementation)

- **Share idempotency state across transports.** Give `commander` and `controlclient` a single shared `Executor` (or a shared `outcomeCache` keyed by command id) so a command executed on either path is not re-run by the other. Cleanest, but threads shared state through the agent wiring; note it is per-agent ephemeral (lost on restart), consistent with today's cache.
- **Make `kill_process` idempotent.** Treat `ESRCH` ("no such process") as `completed` rather than `failed`, since the operator intent ("ensure this PID is dead") is satisfied. Smaller change and independently improves the poll path, but it changes observable command semantics, so it needs an OpenSpec delta and a test update.

Consider also keying the outcome cache by `(host_id, command_id)` if command ids are not globally unique across hosts (Qodo raised the same point in #556).

## Acceptance

[...]

PR #560, opened 2026-06-30
feat(agent): durable cross-transport command dedup ledger (#558)

getvictor:
The push and poll paths each built their own command executor, and the only dedup state was an in-memory cache in the control client. So "a command's side effect runs at most once" held only within the push transport, within one process run. Two windows escaped it:

- Cross-transport: a kill_process executed over the control connection whose outcome didn't reach the server before the stream dropped stays pending; the poll path resumes and re-executes it with a fresh executor, flipping a succeeded kill to failed.
- Restart + PID reuse: the in-memory cache is lost on restart, so a re-delivered kill_process re-executes; since it carries only a bare PID and PIDs are reused, it could signal an unrelated process.

Add a small persistent SQLite ledger (agent/commandledger), keyed by command id, that both transports consult through the shared commander.Executor. The executor write-ahead claims a command (status "executing") before its side effect and records the terminal outcome after; on a re-delivery it replays a recorded terminal outcome instead of re-running, and refuses to re-run a command left in "executing" by a crashed prior attempt (so it never re-signals a reused PID). The control client's in-memory cache is removed (subsumed by the ledger). One ledger is opened in cmd/fleet-edr-agent and shared by both paths; it is pruned on the existing hourly loop.

Includes the openspec delta (durable-command-dedup) with a new agent-command-executor requirement + scenarios (cross-transport and restart), each carrying a test marker.

## Checklist

      (flagged below).

## Summary by CodeRabbit

* **New Features**
  * Command handling now keeps a durable record of outcomes, helping prevent duplicate command execution across reconnects and restarts.
  * Commands delivered through different paths now share the same deduplication behavior.

* **Bug Fixes**
  * Reduced the chance of non-idempotent commands running more than once after connection drops or agent restarts.
  * Improved reliability when a command was already in progress or had previously completed.

### DIFF
--- a/agent/cmd/fleet-edr-agent/main.go
+++ b/agent/cmd/fleet-edr-agent/main.go
@@ -515,8 +515,10 @@ func pruneLoop(ctx context.Context, q *queue.Queue, ledger *commandledger.Store,
 			case pruned > 0:
 				logger.InfoContext(ctx, "pruned old events", "count", pruned)
 			}
-			// Bound the command ledger too: a command-outcome row older than the prune age is far past any plausible re-delivery.
-			if n, err := ledger.Prune(ctx, pruneAge); err != nil {
+			// Bound the command ledger on its own, longer retention: a command stays eligible for re-delivery until it leaves the
+			// server's pending state, so its dedup row must outlive any realistic offline-then-reconnect window (not the 24h event age),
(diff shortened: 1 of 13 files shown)


getvictor:
@coderabbitai full review

PR #562, opened 2026-06-30
fix(detection): wire SetMetrics before Run in tests to close the data race (#561)

### LINKED ISSUE #561: Data race: detection pipeline SetMetrics vs running loops (intermittent -race CI failure)
## Summary

The detection pipeline's `MetricsRecorder` field is set after construction (two-phase wiring) without synchronization, while a runner goroutine reads it. `go test -race` intermittently detects a data race. It is pre-existing and unrelated to whatever PR happens to catch it (it surfaced on PR #560, whose change is agent-only; #559's CI was green on identical detection code).

## Evidence

CI "Server tests" (with `-race`) on PR #560 failed:

```
--- FAIL: TestBootstrap_IntakeModeIsNoOp        race detected during execution of test
--- FAIL: TestSetMetrics_PropagatesToEngineAndIntake   race detected during execution of test
```

The race stack points at `server/detection/internal/pipeline/queueprune.go:73`:

```go
func (r *QueuePruneRunner) Run(ctx context.Context) (int64, error) {
    ...
    if r.metrics != nil {            // <- read, in the Loop goroutine
        r.metrics.QueueRowsPruned(ctx, pruned)
    }
```

racing `QueuePruneRunner.SetMetrics` (writes `r.metrics`, no lock), reached via `Detection.SetMetrics` (`server/detection/bootstrap/bootstrap.go`). The test calls `SetMetrics` after `newDetection(ModeFull)` has launched `d.Run` (which starts the queue-prune `Loop` immediately), so the write and the read overlap.

It is intermittent: it did not reproduce in 10 local `-race` runs (`-count=2` then `-count=8`), but CI caught it.

The same unsynchronized `SetMetrics`-vs-loop pattern likely exists for the other components `Detection.SetMetrics` fans out to (engine, intake, retention, processttl).

## Fix options

- **Synchronize the field**: store the recorder in an `atomic.Pointer` (or guard with a mutex) in each runner that has a post-construction `SetMetrics` and reads `metrics` from a loop. Smallest behavior change; makes the two-phase setter race-safe by contract.
- **Order it instead**: ensure `SetMetrics` is always called before the runner's loop starts (construction-time wiring), and make the test follow suit, so there is no concurrent access. Removes the need for synchronization but constrains the wiring.

## Acceptance

- `go test -race -tags integration ./server/detection/...` is race-free across repeated runs (e.g. `-count=20`).
- The two-phase `SetMetrics` is documented as safe to call concurrently with the running loops (if the synchronize option is taken), or the wiring guarantees it is never concurrent (if the ordering option is taken).

getvictor:
Closes #561.

## What it was

An intermittent `-race` failure in CI. The detection setters (`SetMetrics`, `LoadActive`, `SetModeResolver`) are set-once construction-phase config the running loops read. Production is safe: cmd/main wires them before `go runDetection`, so the loops observe them via the goroutine-launch happens-before. The race was test-induced: the `newDetection(ModeFull)` helper launches `d.Run` during construction, and `TestSetMetrics_PropagatesToEngineAndIntake` then called `d.SetMetrics` afterward, racing the pipeline runners' immediate first sweep (which reads the metrics recorder).

## The fix

Make the test honor the same wire-before-`Run` order production uses: `detectionOpts` gains `metrics` and `ruleProvider`, which the helper installs before launching the loops; the test passes them via opts instead of calling the setters after `newDetection`. The contract ("call before Run, not concurrently with the loops") is now documented on all three setters.

No production logic changed. Locks were deliberately not added to the metrics/rules fields: they are set-once-before-launch config published via the `go` happens-before edge, so synchronizing would put a mutex on hot-path reads for data that never changes after startup and mis-model a one-shot construction setter.

## Verification

- The two flagged tests at `-race -count=20`, and the full detection integration package at `-race -count=2`.
- build + `go vet -tags integration`, golangci `0 issues`, new-code gate `0 issues`, dashes clean.

https://claude.ai/code/session_01BxbWcMpjYFGpHZDU2d2mfm

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup reliability for detection by ensuring configuration is applied before processing begins, reducing the risk of early-run race conditions.
  * Fixed test coverage to reflect the safer initialization flow and validate metrics behavior during startup.

* **Documentation**
  * Clarified setup guidance for detection initialization, including when configuration must be applied relative to runtime.

### DIFF
--- a/server/detection/bootstrap/bootstrap.go
+++ b/server/detection/bootstrap/bootstrap.go
@@ -251,8 +251,11 @@ func (d *Detection) SetMetrics(m api.MetricsRecorder) {
 }
 
 // SetModeResolver wires the per-host rule-mode resolver into the engine AFTER construction (issue #459). cmd/main passes the rules
-// context's detection-config service. No-op in ModeIntake (no engine). Like SetMetrics, it is set-once construction-phase config and
-// MUST be called before Run, not concurrently with the running loops.
+// context's detection-config service. No-op in ModeIntake (no engine). It is set-once construction-phase config, but unlike the metrics
+// recorder (read by the pipeline runners on their immediate first sweep at Run start) the engine reads the mode resolver only while
+// evaluating a claimed event batch. So it need only be wired before events are evaluated, not necessarily before Run: calling it after
+// newDetection but before any events flow is safe (as the integration tests do); what is unsafe is calling it concurrently with the
+// engine's evaluation of a batch.
 func (d *Detection) SetModeResolver(m rulesapi.RuleModeResolver) {
 	if d.engine != nil {
 		d.engine.SetModeResolver(m)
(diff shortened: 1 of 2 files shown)


getvictor:
@coderabbitai full review

PR #559, opened 2026-06-30
Control channel single port

getvictor:
## Checklist

- [ ] **OpenSpec updated for behavior changes.** If this PR changes observable behavior (a detection rule, the event
      wire schema, persistence semantics, an API/wire shape, or an emitted event), `openspec/specs/**` is updated (and
      an `openspec/changes/` proposal added), with a `spec:<id>` test marker for any new/renamed scenario. Only if this
      PR changes NO observable behavior (a comment / refactor / gofmt / dep bump that happens to touch a scoped path)
      may you assert `no-behavior-change` (label or `[no-behavior-change]` in the title) to clear the OpenSpec-sync gate.
      That is an assertion a reviewer will check, not a way to skip the spec for a real behavior change.
- [ ] Tests added/updated for the change (unit / integration / efficacy / UI as applicable).
- [ ] Agent/extension change touching ESF, XPC, or the event wire format: exercised on a live macOS VM before RC
      (flagged below).
- [ ] **Docs updated for user-facing changes.** If this PR changes a user-facing surface (the React UI, an HTTP
      handler, or a detection rule), `docs/` or `CHANGELOG.md` is updated in the same PR. Only if nothing a user sees
      changed (an internal refactor, a non-visible UI tweak) may you assert `no-docs-change` (label or
      `[no-docs-change]` in the title) to clear the Docs-sync gate. The model is in `docs/doc-versioning.md`.

## Summary by CodeRabbit

* **New Features**
  * The app now uses a single shared port for the server UI/API and the agent control channel.
  * The agent automatically derives its control endpoint from the main server URL, using secure or insecure transport based on the URL scheme.

* **Bug Fixes**
  * Improved control-channel shutdown so active streams close cleanly during server stop.
  * Updated proxy and HTTP/2 handling so long-lived control connections work reliably through the shared listener.

Issue #557, opened 2026-06-30
Re-test agent/extension changes on the notarized v0.4.0 RC (enforcement paths can't be validated on dev VMs)


getvictor:
## Summary

Before closing v0.4.0, re-run real-tool QA on the **notarized RC build** for every merged change that touches the **agent** or the **system / network extensions** (including the agent↔extension XPC / wire format and ESF/NE behavior). Dev VMs cannot validate the enforcement paths, so a green dev-VM check is not sufficient sign-off for these changes.

## Why this is needed (dev VMs can't validate enforcement)

The `edr-dev` VM runs **ad-hoc-signed, un-notarized** extensions for fast iteration. Endpoint Security treats an ad-hoc/un-notarized ES client specially: it **redacts exec-target fields and forces `is_platform_binary = true` on every exec** (see the #187 note in `extension/edr/extension/ESFSubscriber+AuthExec.swift`). The AUTH_EXEC handler has a platform-binary carve-out that immediately `ES_AUTH_RESULT_ALLOW`s anything flagged platform (`ESFSubscriber.swift`, #205). Net effect: on a dev VM, **application-control never reaches a rule decision** — every exec is allowed before any rule is consulted, regardless of rule type (CDHASH / SIGNINGID / TEAMID / …).

So the *delivery* and *apply* paths are testable on dev (control channel → agent → XPC → "applied app control snapshot"), but the actual **exec-deny enforcement** is only observable on a **notarized** extension — i.e., on `edr-qa` running the signed RC. This generalizes: any agent/extension behavior that depends on real signing, ESF target fidelity, or enforcement must be re-validated on the RC.

## What to do once the RC is cut

For each v0.4 change touching the agent or extensions, re-run the relevant real-tool QA on `edr-qa` with the notarized RC, and record the result on the change's issue/PR. Categories to cover:

- Agent↔extension **XPC** connect + hello/peer-requirement (production signing, not the dev ad-hoc workaround).
- **ESF event delivery** fidelity (real `team_id` / `signing_id` / `is_platform_binary`, not the dev-redacted values).
- **Application-control enforcement** (AUTH_EXEC deny for a matching rule; rule-type matching that depends on un-redacted target fields).
- **Network extension** filter / DNS-proxy behavior.
- **Command delivery** over the agent (poll and the new push control channel) end to end.
- Reboot / sysext-upgrade / version-skew behavior on a signed install.

## Concrete deferred items so far

[...]

PR #564, opened 2026-07-01
Close #477: control-channel closeout tests + fix the shared-listener stream flap

### LINKED ISSUE #477: Push-based agent control channel: replace GET /api/commands polling (split from #405)

getvictor:
## Summary

Closes out the push-based agent control channel (#477). Most of the feature already merged (#552/#553/#556/#559); this PR adds the two remaining test items, reconciles the openspec task checkboxes with what shipped, and fixes a real defect that the live edr-dev QA surfaced.

## What's here

**Closeout tests + reconciliation (`4238ef91`)**
- `TestControlGatewayPushLifecycle_RealMySQL` (task 5.1): drives a command through the real response service against MySQL over the production gRPC-over-h2c `ServeHTTP` path; asserts the `commands` row walks pending -> acked -> completed with ack/complete timestamps and result persisted, matching the poll path. The prior `TestBuildControlGateway` only covered wiring.
- `TestGatewayLiveness` (task 5.7): connection presence advances last-seen via the injected heartbeat with no command poll or telemetry upload, and the bumps stop once the connection drops so the host ages to offline.
- Checked `push-command-channel` tasks 1.1-6.1 (all verified in tree) and `control-channel-single-port` task 4.1 (Caddy h2c already in `packaging/caddy`).

**Bug fix from live QA (`b1a9e1b1`) — production behavior change**
- Live QA on edr-dev showed the control stream reconnecting every ~10s with `i/o timeout` (visible in SigNoz as a run of 10s `ControlChannel/Connect` spans) instead of holding one persistent connection.
- Root cause: the single-port merge (#559) serves the control gRPC gateway on the shared REST `http.Server`, whose `ReadTimeout` (10s) and `WriteTimeout` (30s) bound the WHOLE HTTP/2 stream, not one message. Commands still flowed (reconnect + at-least-once re-delivery), but idle hosts churned reconnects instead of holding a connection, defeating a #477 acceptance goal.
- Fix: `mountControlChannel` clears the per-stream read/write deadlines for the `application/grpc` branch via `http.NewResponseController`; the REST/UI surface keeps its timeouts. Degrades safely (reconnect + retained short-poll) if a transport refuses the clear.
- Delta: `control-channel-single-port` gains task 2.6, a "Control stream not bounded by REST timeouts" scenario, and a design note.

## Live QA (edr-dev, task 6.1) — passed after the fix

- **Sub-second delivery**: `POST /api/commands` kill delivered pending -> acked -> completed in ~130-150ms end to end; target processes actually killed.
- **Stable connection post-fix**: zero flaps over a 40s window; SigNoz shows one 114s `ControlChannel/Connect` span (clean close) vs the pre-fix 10s `i/o timeout` spans.
- **Offline/online flip**: `last_seen` ages monotonically on disconnect (toward the 5-min offline cutoff) and resets on reconnect, driven purely by the control connection.
- **Trace continuity**: the `ControlChannel/Connect` span (otelgrpc) carries child SQL spans (outcome `UpdateStatus`, liveness, delivery queries).

## Verification

Local gates green: gateway + response-integration + httpserver tests, `go vet -tags integration`, custom golangci-lint, dash-lint, `openspec validate --strict`, `spectrace check --strict` (546/546). The regression test was confirmed to fail (`RST_STREAM`) without the fix.

## Not in scope / still open for #477

- #557 (RC validation of the proxied stream on the notarized build).
- Release-time archive of the two `openspec/changes/` folders.

## Summary by CodeRabbit

* **New Features**
  * Long-lived control-channel streams can now share the same port as the REST/UI server without being cut off by regular server timeouts.
  * Control traffic continues to use the shared endpoint, with improved handling for streaming connections.

* **Bug Fixes**
  * Fixed an issue where control streams could disconnect periodically during normal operation.
  * Added graceful fallback behavior if timeout clearing is not available.

* **Tests**
  * Added regression and integration coverage for stream longevity, liveness, and end-to-end command delivery.

### DIFF
--- a/openspec/changes/control-channel-single-port/design.md
+++ b/openspec/changes/control-channel-single-port/design.md
@@ -16,6 +16,10 @@ The first implementation used `cmux` to split gRPC from REST by sniffing the con
 
 Because the gateway is served via `grpc.Server.ServeHTTP` (riding net/http's HTTP/2 server), `grpc.Server.GracefulStop` is unusable: its `serverHandlerTransport` has no `Drain`, so GracefulStop panics. Shutdown is instead driven by `http.Server.Shutdown`, with the gateway cancelling its live streams so their handlers return promptly: `Gateway.Stop` marks the gateway closing (a connection accepted mid-shutdown is rejected) and cancels every registered connection's context. The control streams are long-lived and would otherwise hold `http.Server.Shutdown` open until its deadline. The agent's keep-alive PINGs need no special server policy: net/http's HTTP/2 server answers them without gRPC's strict ping-flood GOAWAY, so the previous keepalive-enforcement option is dropped.
 
+## Per-stream deadlines cleared for the control channel
+
+Sharing the REST `http.Server` means the control stream inherits its per-request `ReadTimeout` and `WriteTimeout`. On an HTTP/2 stream those bound the whole request/response, not one message, so a 10s `ReadTimeout` tears the long-lived stream down every 10s (observed in live QA as a ~10s reconnect flap, and in SigNoz as a run of 10s `ControlChannel/Connect` spans ending in `i/o timeout`). The multiplexer clears both deadlines for the `application/grpc` branch via `http.NewResponseController` (`SetReadDeadline`/`SetWriteDeadline` to the zero time) before delegating to the gateway; the REST and UI surface keeps its timeouts. Delivery still degrades safely if a transport ever refuses the clear: the agent reconnects and the retained short-poll remains the floor, so no command is lost.
+
 ## Deployment
 
 In the quickstart / single-VM model a front proxy (Caddy) terminates the public TLS and reverse-proxies to the server's plaintext port. The proxy SHALL forward to the upstream over HTTP/2 cleartext (h2c) so the bidirectional control stream is carried, not just HTTP/1.1 REST. The server's plaintext mode accepts h2c for exactly this.
--- a/openspec/changes/control-channel-single-port/specs/server-configuration/spec.md
+++ b/openspec/changes/control-channel-single-port/specs/server-configuration/spec.md
@@ -2,7 +2,7 @@
 
 ### Requirement: The agent control channel shares the main server listener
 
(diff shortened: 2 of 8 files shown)


getvictor:
@coderabbitai review

PR #571, opened 2026-07-01
feat(detection): reconcile exclusion match types with rules + signature exclusions for suspicious_exec (#520)

### LINKED ISSUE #520: Detection exclusions: reconcile match types with rules (UI offers all, rules consult few) and add signature-based exclusions to suspicious_exec
## Summary

The detection-config exclusion editor (UI + API) presents every exclusion match type as valid for every rule, but each rule only consults a fixed subset of match types at evaluation time. An operator can create an exclusion whose `(rule_id, match_type)` pair is never read, and it is accepted, stored, and shown as active while doing nothing. This is both a correctness/UX defect (silently dead config) and the blocker for the motivating use case: excluding a code-signed developer tool (for example Claude Code) by its signing identity instead of a spoofable path glob.

### 1. Match types and rules are not reconciled

There are 8 exclusion match types defined (`server/rules/api/detectionconfig.go`), but only 3 distinct ones are consulted by any rule, and only 4 of the 10 catalog rules consult exclusions at all. The UI lists all 8 for every rule (`ui/src/components/DetectionConfig/DetectionConfig.tsx:23-31`, rendered unconditionally at line 267), and the API only validates that the match type is a valid enum value and that `rule_id` is non-empty (`server/rules/internal/detectionconfig/store.go:138`). It does not validate that the match type is one the rule consumes, nor even that `rule_id` names a real rule.

Coverage today (match type actually consulted per rule, from the `.Excluded(...)` call sites):

| Rule | Match type(s) actually consulted |
| --- | --- |
| `suspicious_exec` | `parent_path_glob` |
| `privilege_launchd_plist_write` | `team_id` |
| `persistence_launchagent` | `path_glob` |
| `sudoers_tamper` | `path_glob` |
| `credential_keychain_dump` | none |
| `application_control_block` | none |
| `dyld_insert` | none |
| `dns_c2_beacon` | none |
| `osascript_network_exec` | none |
| `shell_from_office` | none |

Match types and whether any rule consumes them:

| Match type | Consumed by |
| --- | --- |
| `path_glob` | `persistence_launchagent`, `sudoers_tamper` |
| `parent_path_glob` | `suspicious_exec` |
| `team_id` | `privilege_launchd_plist_write` |
| `signing_id` | nothing |
| `cdhash` | nothing |
| `sha256` | nothing |
| `command_substring` | nothing |
| `domain` | nothing |

So 5 of 8 match types are dead catalog-wide, 6 of 10 rules accept no effective exclusion, and of the 80 (rule × match type) combinations the UI offers, only 4 do anything.

### 2. `suspicious_exec` can only exclude by path, which is the spoofable option

`suspicious_exec` consults only `parent_path_glob` (`suspicious_exec.go:515`), and its `execPayload` does not even unmarshal the `code_signing` field that is present on the exec event. Because a path glob's `*` matches `/`, a leading-wildcard exclusion like `*/claude/versions/*` matches that fragment anywhere on disk, so an attacker who can write to `/tmp` creates `/tmp/claude/versions/payload`, runs it from a shell, and lands inside the exclusion. There is no way today to exclude a benign parent by its code signature.

Motivating example: the Claude Code binary on a developer host is Developer-ID signed (`TeamIdentifier=Q6L2SF6YDW`, `Identifier=com.anthropic.claude-code`). The correct, update-stable, non-spoofable exclusion is `team_id=Q6L2SF6YDW`, but `suspicious_exec` cannot consult it.

The data is already there: `Process.CodeSigning` and `Process.CDHash` are persisted on the process record (`server/detection/api/types.go`), fed by the `code_signing` field the agent already puts on exec events. The rule just does not read it. Extending `suspicious_exec.parentExcluded()` to also match `team_id` / `signing_id` / `cdhash` against the parent process's already-stored signing needs no agent or wire change.

## Proposed direction

Pick one consistent model and make the UI, API, and rules agree on it. The recommended shape:

1. Each rule declares the match types it supports as a single source of truth (for example a `SupportedExclusionMatchTypes() []ExclusionMatchType` on the rule interface), surfaced in `GET /api/rules`.
2. The create-exclusion API validates `match_type` against the target rule's supported set and rejects unsupported pairs with a clear error. It should also validate that `rule_id` names a real rule (it does not today).
3. The UI fetches the supported match types for the selected rule and only offers those in the match-type dropdown.
4. Extend `suspicious_exec` to support signature-based exclusions (`team_id`, `signing_id`, `cdhash`) on the parent, reading the already-persisted `Process.CodeSigning` / `Process.CDHash`. This is the concrete win that lets `team_id=Q6L2SF6YDW` replace the spoofable path glob for Claude Code.
5. Audit every other rule and decide, per rule, which match types are meaningful, then either implement them or ensure they are not offered. Candidates worth considering: `domain` for `dns_c2_beacon`, `command_substring` for argv-inspecting rules, and signing-based match types for the exec/persistence rules that currently only do path.

## Notes for the implementer

- This changes exclusion matching semantics, API validation, and the rule capability surface, so it ships an `openspec/changes/<name>/` delta (proposal + tasks + spec delta) in the same PR.
- Consider existing stored exclusions whose `(rule_id, match_type)` pair would become invalid under the new validation: decide whether to migrate, flag, or leave them inert, and surface them in the UI rather than silently dropping.
- Add a test asserting that for every rule, every match type the API/UI offers is one the rule actually consults (close the drift permanently).

getvictor:
## Summary

Fixes #520. Reconciles the detection-config exclusion editor (UI + API) with what each rule actually consults, and adds non-spoofable signature-based exclusions to `suspicious_exec`.

Before this change the editor offered all 8 `ExclusionMatchType` values for every one of the 10 rules, but only 4 rules consult exclusions and only 3 match types are ever read. An operator could store a `(rule_id, match_type)` pair no rule reads: accepted, persisted, shown active, suppressing nothing. `suspicious_exec` could also only exclude by parent path, which is spoofable (`*/claude/versions/*` also matches `/tmp/claude/versions/payload`), blocking the motivating use case of excluding a signed developer tool like Claude Code by `team_id=Q6L2SF6YDW`.

## What changed

- **Single source of truth**: `SupportedExclusionMatchTypes()` on the `api.Rule` interface (compile-error gate on all 10 rules), surfaced on `GET /api/rules`.
- **API validation**: creating an exclusion now rejects an unknown `rule_id` and a `match_type` the rule does not consult, returning HTTP 400 with a message naming the supported set.
- **UI**: the exclusion editor offers only the selected rule's supported match types and resets the selection when the rule changes.
- **`suspicious_exec` signature exclusions**: suppresses by the non-shell parent's already-persisted code-signing identity (`team_id` / `signing_id` / `cdhash`) in addition to `parent_path_glob`. No agent or event-wire change. An unsigned lookalike is not suppressed.
- No new exclusion capabilities added to other rules; rules that consult nothing offer nothing.

Existing stored exclusions whose pair is now unsupported are already inert and are left in place; validation applies to new creates only (no migration).

## Behavior change

Creating an exclusion for an unsupported `(rule_id, match_type)` pair or an unknown rule now returns HTTP 400 where it previously returned success. This is the intended fix and is captured in the OpenSpec delta.

## Tests

- Anti-drift guard pinning each rule's declared set + a recording resolver proving `suspicious_exec` queries exactly what it declares.
- `suspicious_exec` signature-suppression (team_id / signing_id / cdhash) and unsigned-lookalike-not-suppressed.
- Service validation (unsupported pair, unknown rule, no-exclusion rule, unset-map skip).
- Handler: `/api/rules` includes the field; the 400 rejection message reaches the body.
- UI: picker offers only the supported set and resets on rule change.
- Ships `openspec/changes/reconcile-exclusion-match-types` (proposal, tasks, spec delta); spectrace markers reference the new scenarios.

Verified: `go build`, `go vet -tags integration`, `go test ./server/rules/... ./server/detection/...`, full UI suite (435 pass), `task lint:go` / `lint:dashes` / `lint:md:prose`, `openspec validate --all --strict`, `spectrace check --strict`.

## Summary by CodeRabbit

* **New Features**
  * Exclusion options now adapt to the selected rule, showing only supported match types.
  * `suspicious_exec` now supports additional signature-based parent exclusion matching.

* **Bug Fixes**
  * Prevented saving exclusions with unsupported rule/match-type combinations.
  * Reset match type selection when changing rules to avoid stale values.

* **Chores**
  * Rule listings now include supported exclusion match types for a more accurate admin experience.

### DIFF
--- a/docs/recommended-exclusions.md
+++ b/docs/recommended-exclusions.md
@@ -9,21 +9,26 @@ Add these on the detection-config exclusions surface in the console (sign in as 
 An exclusion is `(rule_id, match_type, value)`. The value is matched per the rule's match type:
 
 - `parent_path_glob`: matches a chain's non-shell parent path. `*` matches any run of characters including `/`. A pattern with no `*` is an exact match.
-- `team_id`: matches an Apple Developer team ID exactly.
+- `team_id`: matches an Apple Developer team ID exactly (the `TeamIdentifier` field of the code signature).
+- `signing_id`: matches a code-signing identifier exactly (the `Identifier` field, for example `com.anthropic.claude-code`).
+- `cdhash`: matches a binary's code-directory hash exactly (40 lowercase hex characters), pinning one exact build.
 - `path_glob`: matches an absolute filesystem path with the same glob semantics as `parent_path_glob`.
 
-Which rule consumes which match type is fixed by the rule:
+Which rule consumes which match type is fixed by the rule, and the console offers only the match types the selected rule actually consults. Creating an exclusion for a `(rule_id, match_type)` pair the rule does not consult, or for a rule that does not exist, is rejected by the API, so a stored exclusion can no longer silently do nothing:
 
-| Rule                            | Match type used    |
-| ------------------------------- | ------------------ |
-| `suspicious_exec`               | `parent_path_glob` |
-| `privilege_launchd_plist_write` | `team_id`          |
-| `persistence_launchagent`       | `path_glob`        |
-| `sudoers_tamper`                | `path_glob`        |
+| Rule                            | Match types used                                      |
+| ------------------------------- | ----------------------------------------------------- |
+| `suspicious_exec`               | `parent_path_glob`, `team_id`, `signing_id`, `cdhash` |
+| `privilege_launchd_plist_write` | `team_id`                                             |
+| `persistence_launchagent`       | `path_glob`                                           |
+| `sudoers_tamper`                | `path_glob`                                           |
+
+`suspicious_exec` matches its signature match types (`team_id`, `signing_id`, `cdhash`) against the chain's non-shell parent. A code-signed parent can therefore be excluded by its non-spoofable signing identity instead of a path glob, which is the preferred form for any signed tool (see the caveats below).
 
 ## Caveats before you add an exclusion
 
-- A `suspicious_exec` `parent_path_glob` exclusion silences BOTH arms of the rule for that parent (the temp-path exec arm and the outbound-network arm). You are trusting everything that parent does, not just the one connection you saw.
+- Prefer a signature exclusion over a path glob whenever the parent is code-signed. A `team_id` (or `signing_id` / `cdhash`) exclusion on `suspicious_exec` matches the parent's signing identity, which survives version churn and cannot be spoofed by planting a binary at a lookalike path, unlike a `parent_path_glob`. Reach for a path glob only when the tool is unsigned.
+- A `suspicious_exec` exclusion silences BOTH arms of the rule for that parent (the temp-path exec arm and the outbound-network arm), whichever match type you use. You are trusting everything that parent does, not just the one connection you saw.
 - Never start a path glob with `*`, and minimize interior wildcards. Because `*` matches any run of characters including `/`, a leading-wildcard pattern like `*/claude/versions/*` matches that fragment anywhere on disk, so an attacker who can write to `/tmp` creates `/tmp/claude/versions/payload` and runs it to land inside the exclusion. Anchor to the full absolute install path instead; the longer the literal prefix, the less an attacker can spoof.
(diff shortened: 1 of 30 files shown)


getvictor:
@coderabbitai review

Issue #423, opened 2026-06-17
Hunting: honest tree truncation + result metadata


getvictor:
Part of #415 · Phase 0 · Size **S** · depends on: none

### Goal

Stop silently dropping process nodes. Return result metadata so the UI can tell the analyst the tree is truncated.

### Why

`GetProcessTree` applies `LIMIT 2000` server-side and the UI renders whatever comes back with no indicator. On a busy host a 15-minute window matched 2,588 rows, so the analyst sees 2,000 with no signal that ~600 are missing. This is a correctness/trust bug and it blocks honest messaging in every later story.

### Scope

- **server**: `TreeResponse` gains `total_matched` (count of all rows matching the window predicate) and `truncated` (bool). Add the `COUNT(*)` companion to the existing window predicate in `server/detection/internal/mysql/processes.go`.
- **UI**: `TreeResponse` type + a banner in `ProcessTree.tsx`: "Showing 2,000 of 2,588 processes. Narrow the time range or use search."

### Acceptance criteria

- API response includes `total_matched` and `truncated`.
- When `truncated` is true the UI shows the count banner; when false it does not.
- `total_matched` equals the true overlap count, independent of the `limit`.

### Tests

- Handler test: window with > limit rows sets `truncated: true` and the correct `total_matched`.
- UI test: banner renders only when `truncated`.

Issue #421, opened 2026-06-17
Hunting: lazy / progressive tree expansion + virtualization


getvictor:
Part of #415 · Phase 3 · Size **M** · depends on: #415 story 2

### Goal

Open the tree collapsed at the roots and fetch children on demand; virtualize rendering so only on-screen nodes lay out. This is what lets the tree survive 6h/24h windows.

### Why

Today the UI fetches up to 2,000 rows and D3 lays out every one up front. Progressive disclosure (the CrowdStrike/SentinelOne default) plus virtualization removes both the payload and the layout-cost ceilings, and makes the truncation cap largely moot for interactive use.

### Scope

- **server**: support root-level fetch + expand-on-demand. The child query already exists (`GetChildProcesses` in `server/detection/internal/mysql/processes.go`); wire it to an endpoint or extend the tree endpoint with a `parent`/depth param.
- **UI**: roots collapsed by default; expanding a node lazy-loads its (aggregated) children; virtualize the D3/DOM render to on-screen nodes.

### Acceptance criteria

- Initial load fetches only roots + first level, not the whole forest.
- Expanding a node fetches its children once and caches them.
- Large windows stay interactive (no full-forest layout).

### Tests

- Handler test for the children-of-parent fetch within a window.
- UI: expand triggers exactly one child fetch and renders results; collapse retains cache.

Issue #418, opened 2026-06-17
Hunting: process search endpoint (filters + cursor pagination + sort)


getvictor:
Part of #415 · Phase 2 · Size **L** · depends on: #415 story 1

### Goal

A server-side process search endpoint with filters, cursor pagination, and sort. This is where high-volume hunting actually happens: replace "render 2,000 nodes" with "query 12 rows".

### Why

The tree is for one scoped story; the table is for volume. Across a 1h/6h window there are tens of thousands of processes that no graph can show. Analysts need to ask `path = grep`, `unsigned AND has_network`, `uid = 0 AND time in [t0,t1]` and get a precise, paginated answer. This story also absorbs the retired "benign suppression" residual: cap-aware, server-side filtering (and optional signature-based "signed Apple platform" exclusion) lives here, where filtering and the row budget are decided together.

### Scope

- **server (detection context)**: `GET /api/hosts/{host_id}/processes` with query params for `path` (prefix/contains), `signing` state, `uid`, `time` range, `has_network`, `exit_reason`, plus `sort` and an opaque `cursor` for keyset pagination. Returns rows + `next_cursor` + `total_matched`. New store query alongside `GetProcessTree`.
- No UI in this story (UI is #415 story 5).

### Acceptance criteria

- Filters compose (AND) and are applied in SQL, not in Go.
- Keyset pagination is stable under concurrent inserts (no row skips/dupes across pages).
- `total_matched` returned for the filtered set.

### Tests

- **PBT**: paginating the full result set with any page size yields exactly the unpaginated set, in order, no dupes.
- Handler tests for each filter and an empty result.

getvictor:
Merged into #582 (epic #577): the endpoint ships fleet-wide from the start with host_id as an optional filter instead of a per-host route, keeping this issue's full contract (SQL-composed filters, keyset cursor pagination, total_matched, benign-suppression residual, PBT pagination invariant). #582 is the source of truth going forward.

Issue #415, opened 2026-06-17
Epic: best-in-class hunting and investigation experience


getvictor:
## Epic: best-in-class hunting and investigation experience

> **Scope note (2026-07-02):** the hunting-table and time-navigation stories moved to epic #577 during the duplicate-merge pass: #418 + #419 were merged into #582 (fleet-wide artifact search + bridge pivots, with `host_id` as an optional filter) and #420 into #581 (compact time control + histogram scrubber). This epic remains the source of truth for the tree-scale stories below; #577 is the source of truth for search, time navigation, and the rest of the investigation UX.

### Problem

The only investigation surface today is **HostList → ProcessTree** (`/hosts/:hostId`) plus a flat **AlertList**. The process graph renders every node in D3 and is capped silently at 2,000 rows server-side (`server/detection/internal/mysql/processes.go` `GetProcessTree`, `LIMIT`). There is no search, no aggregation, no histogram, and no server-side filtering.

Measured on a live host (`B7D79E29-...`, OVH quickstart deployment) for a single **15-minute** window:

- **2,588** processes overlapped the window; the graph silently truncates to 2,000.
- **2,532** were exited (grey), **56** running (green).
- The volume is legitimate dev-machine churn: `grep ×1000`, `docker ×270`, `jspawnhelper ×240`, `ssh ×170`, `git ×129`, plus the Go toolchain (`compile`/`link`/`asm`), `clang`, `node`, Chrome helpers, Spotlight.

A raw 2,588-node graph for 15 minutes (and far more for 1h/6h) is not a graph problem, it is a **search-and-aggregate** problem. Mature EDRs (CrowdStrike Falcon, SentinelOne Storyline, Microsoft Defender) reserve the tree for a single scoped story and do high-volume hunting in a filtered, paginated table with a time histogram on top.

### Goal

Move the hunting experience to best-in-class by (a) shrinking the tree to meaningful nodes via aggregation and signal-first defaults, (b) adding a server-side search + table for raw volume (now #582 in epic #577), and (c) adding temporal navigation (now #581 in epic #577) and a scoped alert-anchored investigation view.

### Note on the existing "Show system" toggle

[...]

PR #574, opened 2026-07-02
feat(detection,ui): server-side sibling aggregation for the process tree (#416)

### LINKED ISSUE #416: Hunting: server-side sibling aggregation (collapse repeated execs)
Part of #415 · Phase 0 · Size **M** · depends on: #415 story 1

### Goal

Collapse repeated identical child executions under the same parent into a single node carrying a count and a min/max time range. `grep ×1000` becomes one node, not a thousand dots.

### Why

The dominant volume on a real host is repetition: `grep ×1000`, `docker ×270`, `jspawnhelper ×240` in a single 15-minute window. None of it is system-path noise (so the `showSystem` toggle does not touch it). Aggregating siblings is the single biggest "manageable" win, and doing it server-side shrinks the payload itself rather than just the render, which also relieves the truncation cap.

### Scope

- **server**: in the forest/query layer (`server/detection/internal/graph` + `mysql`), group children of the same parent by `path` (and signing identity) into an aggregated node `{ path, count, first_fork_ns, last_fork_ns, exited_count, running_count }`. Aggregated nodes are expandable on demand (defer the row fetch to the lazy-expand story #7, or return a capped sample).
- **UI**: render an aggregated node with a `×N` badge; clicking expands to the underlying executions (or the sample).

### Acceptance criteria

- A parent that spawned N identical-path children renders as one node with `×N`.
- Counts split exited vs running; the time range reflects first/last fork.
- Aggregation is opt-outable (a "flatten" toggle) for analysts who want every node.

### Tests

- **PBT** (`pgregory.net/rapid`): for any random fork batch, the aggregated output preserves total leaf count (sum of `count` == input children), preserves order by first_fork, and no child is lost or duplicated.
- Handler test pinning the aggregated node shape.

getvictor:
## Summary

Closes #416 (part of the hunting epic #415, Phase 0). Collapses repeated identical-path **leaf** siblings under the same parent into a single aggregated `×N` node carrying a count, an exited/running split, and a fork-time span, so a busy host's `grep ×1000` / `jspawnhelper ×240` churn renders as a handful of nodes instead of thousands of dots. Aggregation is a pure, order-preserving, lossless read-time transform over the built forest (never drops or duplicates a process, never folds a node that has its own subtree), on by default and opt-outable with `?flatten=1`.

Doing it server-side shrinks the response payload itself, not just the render.

## What changed

**Server**
- `aggregateSiblings` + helpers in `server/detection/internal/graph/query.go`: recursive, pure, groups leaf siblings by binary identity (`path` + `sha256` + `cdhash`), folds groups of ≥2, orders output by first fork time (deterministic).
- Per-group summary carries full-group count, exited/running split, min/max fork time, and a fork-ordered `sample` capped at 8 so the UI can expand in place before the lazy-expand story (#421) lands.
- `flatten bool` threaded through `api.Service.BuildTree` → `service.Service.BuildTree` → operator handler, read via a new `httpserver.ParseBoolParam` (`?flatten=1` returns the raw forest).
- New `AggregatedSiblings` type + `Aggregated` field on `api.ProcessNode` (`omitempty`, so a non-aggregated node's wire shape is unchanged).

**UI**
- Aggregated nodes render as a `×N` badge (green while any member runs, grey otherwise), expand to their sample on click/chevron, and a persisted **Flatten** toggle refetches the raw forest.

**Tests**
- rapid PBT: leaf-count + per-identity multiset preservation, ordering, per-node consistency.
- Table cases + handler tests pinning the wire shape and `flatten`; 3 UI vitest cases.

**OpenSpec**
- `server-sibling-aggregation` change: deltas for `server-process-graph-builder` (ADDED), `server-rest-api` (MODIFIED), `web-ui` (MODIFIED). `openspec validate --strict` passes.

## Design calls
- **Leaf-only** aggregation: collapses the dominant repetition without risking a dropped subtree before lazy expansion (#421). Repeated non-leaf execs stay individual for now (follow-up).
- **Capped in-payload sample** rather than deferring expansion entirely to #7/#421, so "click to expand" works today.
- Depends on #423 per the epic, but functionally standalone; built on `main`. Aggregation shrinks the payload but does not lift the 2,000-row fetch cap (that's #418/#423).

## Manual QA
Seeded a deterministic host on dev:server and drove it in a real browser:
- Aggregated default: `grep ×10` (8 exited / 2 running), `docker ×2`, `vim` individual, `make → cc` subtree intact.
- Clicking `grep ×10` expands to its 8-member sample in place.
- Flatten toggle refetches with `?flatten=1` → all 16 raw nodes.

## Summary by CodeRabbit

* **New Features**
  * Process trees now group repeated identical leaf siblings into compact `×N` nodes by default.
  * Added a persistent **Flatten** toggle to switch between grouped and raw tree views.
  * Aggregated groups can be expanded in place to preview sample members.

* **Bug Fixes**
  * Improved tree ordering and display of grouped processes for more consistent browsing.
  * Process tree rendering and tests were updated to prevent zoom-related issues in the UI.


getvictor:
@coderabbitai review

PR #576, opened 2026-07-02
fix(detection): retry rule evaluation when the subject process is not yet materialized

getvictor:
## Summary

Fixes the silent alert loss behind the 2026-07-02 nightly demo failure (run 28580179395: the source leg lost exactly the `sudoers_tamper` alert with no error logged; the identical commit passed on re-run, confirming a probabilistic race).

Since intra-replica processor concurrency (#535 / PR #541), two claim batches of the same host can be in flight at once (the same window has always existed across replicas per ADR-0011). Rule evaluation runs after its own batch's process rows flush, but nothing orders it after a concurrent batch's flush. When the claim boundary splits an attack's `exec` from its follow-on event, the rule's `GetProcessByPID` misses, the miss was treated as "no finding", and the event was acked: the alert was gone permanently.

## What changes

- **`rules/api.ErrProcessNotYetMaterialized`** + a catalog helper (`resolveSubjectProcess`): rules that resolve the process an event is ABOUT (the pid in the event's own payload, whose fork/exec is ordered earlier in the same host stream) raise the sentinel on a miss while the event's ingest age is inside a 30s materialization grace window. Past the window the historical silent skip applies, so an orphaned event (pre-capture pid, dropped fork/exec) can never hold its batch in a retry loop. Wired into `sudoers_tamper`, `credential_keychain_dump`, `application_control_block`, `persistence_launchagent`, `dyld_insert`, `shell_from_office`, `osascript_network_exec`, `suspicious_exec`.
- **Engine**: propagates exactly that error class so the processor nacks and re-evaluates the batch (alert dedup makes the re-run idempotent); every other rule failure keeps the log-and-swallow per-rule isolation.
- **Demo seeder**: `verify` now requires a fired alert per woven attack's `ExpectRule` and names the missing rules on timeout, instead of passing on "any detection alert" and deferring the failure to the Playwright smoke.

**Deliberately NOT wired**: ancestor/parent-chain walks (a parent can legitimately predate the capture) and `dns_c2_beacon`'s flow resolution (it resolves before its suspicion gate, so a retryable miss would fire for every outbound connect whose exec the agent dropped under load, turning overload into batch-retry storms). All wired call sites resolve only after a payload-level pre-filter, so the sentinel is raised only by already-rare events.

## Performance

One integer comparison on an already-rare nil-lookup branch; a batch retry only when the race actually fires (resolves on the next poll tick). The grace window self-disables under backlog (events arrive at evaluation already older than it), so a stressed processor keeps pre-change behaviour.

## Spec

Ships `openspec/changes/retry-unmaterialized-process-events`: adds the "Retryable evaluation on unmaterialized subject process" requirement to `server-detection-rules-engine`, with spectrace markers on the new tests.

## Testing

- Rule level: young miss raises the sentinel; past-grace miss skips silently; materialized row fires the finding; grace-window edge cases (zero ingest stamp, exact boundary).
- Engine level: the sentinel propagates out of `Evaluate`; an ordinary rule failure stays swallowed.
- Seeder: per-rule fired set feeds the verify predicate; end-to-end run seeds one alert per expected rule.
- Full `server/rules`, `server/detection`, seeder suites pass; `go vet -tags integration` clean; custom golangci-lint 0 issues; `openspec validate --all --strict` and spectrace 546/546 green.

https://claude.ai/code/session_0194fsN2dtGAs5ysUpxfjMxM

## Summary by CodeRabbit

* **New Features**
  * Detection rules now retry briefly when a related process record is still being created, reducing missed alerts from race conditions.

* **Bug Fixes**
  * Alerts are no longer silently dropped during that short window; they are re-evaluated automatically once the process becomes available.
  * Demo seeding now verifies that all expected detection rules fire, improving end-to-end validation.


getvictor:
@coderabbitai review

PR #609, opened 2026-07-04
fix(detection): retry dns_c2_beacon on unmaterialized flow process

getvictor:
## What

The `Demo nightly` **source** leg intermittently loses the (Critical) `dns_c2_beacon` alert. The seeder fails its verify step with `missing_rules=[dns_c2_beacon]` while the other four detection alerts fire, and the identical commit passes on re-run: a probabilistic race, not a regression.

## Root cause (a server-side detection race, confirmed by tracing the pipeline)

With intra-replica processor concurrency (#535) and cross-replica claimers (ADR-0011), the claim boundary can split the attack's `exec` (which materializes the connecting process's row) from its later `network_connect`, and the two land in batches processed concurrently. When the connect is evaluated **before** the exec's batch commits the process row, `resolveFlowProcess` misses, `evalEvent` returned no finding, the connect was acked, and the alert was **silently and permanently lost**.

A longer wait cannot fix it: a dispatched demo-nightly with a 60s verify window and again with a 120s window both failed with `missing_rules=[dns_c2_beacon]`. Once the miss occurs the alert never arrives, because nothing re-evaluates the connect.

`retry-unmaterialized-process-events` closed this class for subject-process rules but deliberately left `dns_c2_beacon` on silent skip, recording it as accepted residual risk: its flow resolution runs **before** the suspicion gate (the gate reads the process path), so a naive retryable miss would fire for every outbound connect whose exec the agent dropped, risking batch-retry storms. This PR closes that residual risk.

## Fix

- **`dns_c2_beacon.go`**: on a flow-process miss, if the connect's ingest age is inside a new grace window, raise `api.ErrProcessNotYetMaterialized` so the engine propagates it and the processor nacks + re-evaluates the batch once the row lands (dedup keeps the re-run idempotent). Past the grace, or with no ingest stamp, keep the historical silent skip so a genuinely orphaned connect cannot loop.
- **`materialize.go`**: new `flowProcessMaterializationGrace = 5s`, materially tighter than the 30s subject grace, plus a shared `withinGrace` helper. The tight bound is the deliberate answer to the storm concern: flow resolution is reachable by any outbound connect (not a pre-filtered event), so 5s still covers the real race (the row commits within a batch flush, clearing on the next tick) while capping how long a genuinely orphaned connect can nack its batch. The age gate self-disables under backlog.
- **Unchanged**: the claim/ack/nack contract, batch flush, alert schema and dedup, the rule's match logic and suspicion gate, and every wire shape. The engine propagation + processor nack path already exist from the sibling change. The grace is a compiled constant; no new config surface.

## Testing

- New unit tests: a young outbound connect with no flow process row raises the sentinel; a past-grace connect skips silently; a zero ingest stamp (fixture replay) skips silently. Existing fixtures (positive beacon + negatives) and `resolveFlowProcess` precedence tests pass unchanged.
- `go build`, `go vet -tags integration`, gofmt, custom golangci-lint (0 issues), `openspec validate --all --strict` (57/57), `spectrace --strict` (543/543, new markers resolve).
- End-to-end: dispatching `demo-nightly` against this branch to confirm the source leg's seeder + smoke go green (see checks).

## Spec

Ships the `retry-unmaterialized-flow-process` openspec delta (new requirement + two scenarios) and reconciles the sibling `retry-unmaterialized-process-events` delta's now-superseded flow carve-out so the batched canonical archive stays consistent.


getvictor:
Folding this into a single combined PR per request. The dns_c2_beacon retryable-flow-process fix here depends on the graph-builder idempotency fix (otherwise the retry's nack storm inflates the process graph and alerts ~10x), so the two ship together in one PR. Superseded by the combined branch fix-dns-c2-beacon-flake.

PR #611, opened 2026-07-04
fix(detection): idempotent batch re-processing + dns_c2_beacon materialization retry

getvictor:
## Summary

Fixes the intermittent `Demo nightly` **source**-leg failure where the seeder times out with `missing_rules=[dns_c2_beacon]` while the other four detection alerts fire. Root-causing it surfaced a real, two-level bug in the detection pipeline, so this ships both fixes as one unit (the second depends on the first).

## Root cause

1. **`dns_c2_beacon` silently drops its alert on a materialization race.** The rule fires on a `network_connect` and looks backward for the connecting process + the resolving `dns_query`. With concurrent processor workers (#535) and cross-replica claimers (ADR-0011), the connect can be evaluated before the batch carrying its `exec` commits the process row; the rule returns no finding, the event is acked, and the (Critical) alert is **permanently lost** with no re-evaluation. A longer seeder timeout can't help (a 60s and a 120s dispatch both failed the same way).
2. **Batch re-processing was not idempotent.** The obvious fix (make the miss retryable, so the processor nacks and re-evaluates) exposed a latent bug: every fork/exec re-INSERTed a process row (the `processes` table had only a surrogate PK), so a nack/re-claim **duplicated process rows**, which cascaded into **duplicate alerts** (dedup keys on `process_id`). A dispatched run proved it: the retry's nack storm inflated the demo to ~4932 processes / ~59 alerts (`credential_keychain_dump` fired 10x for one woven attack) vs the normal ~400 / ~6, self-inflicting the backlog that kept the flake alive.

## What changes

**Commit 1 - graph-builder idempotency (foundational).** Migration `00011` ties each process row to the event that materialized it: `source_event_id` (+ `UNIQUE(host_id, source_event_id)` backstop), `exec_event_id` (the case-b in-place exec UPDATE), `exit_event_id` (the observed exit). The builder calls `EventAlreadyApplied` and no-ops a replayed event, so re-processing a batch is idempotent. Two supporting correctness fixes: an exit can no longer close a **future-forked** row (`fork_time_ns <= exit ts`), and heartbeat freshness is **monotonic** (`GREATEST`). Columns are nullable and NULL for existing rows, so the ALTER needs no dedup and cannot fail on history; the AUTO_INCREMENT PK is unchanged (the SKIP-LOCKED claim path the reverted 00006 deadlocked is untouched).

**Commit 2 - `dns_c2_beacon` retryable flow-process miss.** On a flow-process miss within a tight `flowProcessMaterializationGrace` (5s, deliberately tighter than the 30s subject grace because flow resolution runs before the suspicion gate), the rule raises `ErrProcessNotYetMaterialized` so the processor nacks and re-evaluates once the row lands. This is now safe precisely because commit 1 made re-processing idempotent (no inflation).

## Testing

- **Double-apply == single-apply** across the crafted state-machine scenarios + a property-based random-sequence variant (real MySQL); existing differential (batched == per-event) and full graph state-machine suites pass with 00011 applied.
- `dns_c2_beacon` retry unit tests (young miss -> retryable, past-grace -> skip, zero-stamp -> skip); full catalog + detection integration + unit suites green.
- `go vet -tags integration`, custom golangci-lint (0 issues), `openspec validate --strict` (59/59), `spectrace --strict` (100%).
- **End-to-end**: dispatching `demo-nightly` against this branch to confirm the seeder is green with no inflation and `dns_c2_beacon` fires reliably across runs (see checks / comments).

## Spec

Ships two openspec deltas: `idempotent-batch-reprocessing` (server-process-graph-builder) and `retry-unmaterialized-flow-process` (server-detection-rules-engine), plus a prose reconciliation of the sibling `retry-unmaterialized-process-events` flow carve-out.

Supersedes #607 (verify-timeout bump, wrong fix) and #609 (retry alone, unsafe without idempotency).

## Summary by CodeRabbit

* **New Features**
  * Added retry handling for DNS beacon evaluations when related process data is not yet available, improving alert reliability during concurrent processing.
  * Process reprocessing is now idempotent, helping preserve a stable process graph across retries and replays.

* **Bug Fixes**
  * Prevented duplicate alerts and duplicate process generations during batch reprocessing.
  * Fixed missed DNS C2 alerts caused by race conditions.
  * Ensured process “last seen” timestamps only move forward and exits do not close the wrong process generation.


getvictor:
@coderabbitai review

getvictor:
@coderabbitai review

getvictor:
@coderabbitai review

PR #610, opened 2026-07-04
feat(detection,visibility): host event timeline endpoint (#583, PR 1/2)

getvictor:
## What

Server half of #583 (epic #577): a host-scoped, time-ordered, merged event query the Timeline tab (PR 2) will render. The host page has a process graph for lineage but no flat, filterable event stream for hunting; every leading EDR ships both over the same data.

`GET /api/hosts/{host_id}/timeline` returns one host's exec, network_connect, and dns_query events interleaved in event-time order (newest first) over an event-time window, filterable by `?type=` (a comma subset of the classes) and a case-insensitive `?text=` payload substring, keyset-paginated with `total_matched`, reusing #582's cursor contract. Gated on process-read scoped to the host. Unknown type or malformed cursor is a 400; an unset archive is a 503.

## How

- The visibility `EventArchive` gains `HostTimeline`. All three classes already live in the ClickHouse archive, so one `event_type IN (...)` query merges them in native time order; no API-layer merge of MySQL processes and ClickHouse events. The window bounds event `timestamp_ns` (what the graph's window means), unlike `SearchEvents` which bounds ingest time, so it is a distinct filter type.
- The newest-first count + keyset page tail is shared by `SearchEvents` and `HostTimeline` (store `pageEventsByKeyset`, fake `pageMatched`) so the two paths cannot drift.
- fork/exit are excluded: exec is the process-start row, fork/exit lineage is the graph's job. The allowlist `{exec, network_connect, dns_query}` lives in one place (`visibility api.TimelineEventTypes` / `IsTimelineEventType`).
- Detection delegates through the same reader-seam pattern as the fleet search (`mysql.Store` -> archive), wired in ModeFull; `detection/api` re-exports `HostTimelineFilter`.
- Rows carry the originating pid so PR 2 can cross-link a row to its graph node.

## Tests

Handler tests drive the real in-memory archive through the endpoint: interleave + fork-exclusion, type filter, text match, stable/complete keyset pagination, host scoping, event-time window, and the unknown-type / bad-window / bad-cursor 400s, 503-unset, 403-denied paths. A real-ClickHouse integration test (`//go:build integration`) mirrors the behavioral cases.

`go vet -tags integration` clean, `task lint:go` clean, spectrace 100% (5 new scenario markers), `openspec validate --all --strict` 58/58.

## Manual QA

Dev server + real ClickHouse, seeded host `AAAA1059` (7 exec + 7 fork + 1 network):

[...]

Issue #583, opened 2026-07-02
Investigation UX: unified host event timeline (exec + network + DNS)


getvictor:
Part of #577 · Wave 2 · Size **L** · follows the endpoint patterns of #582 (which absorbed #418/#419)

### Goal

A flat, time-ordered, filterable event timeline on the host page: exec, network, and DNS rows merged into one stream, as a sibling tab to the graph over the same time window.

### Why

Analysts use the graph for lineage ("how") and a flat list for hunting ("when" + filtering + pivoting); every top product ships both over the same data (SentinelOne's events table beside the graph, MDE's timeline tab, Elastic's event pills). Graphs stop being readable past a few hundred nodes; the filterable list is the fallback that actually scales. #582's table covers processes (plus per-artifact network/DNS lookups); this story unifies event types into one host-scoped investigation stream.

### Scope

- **server**: a host-scoped, time-ordered merged event query (exec + network_connect + dns_query to start; file events when they exist) with type filters, text match, and cursor pagination following the #582 contract. Decide merged-endpoint vs per-type queries merged at the API layer during the openspec proposal.
- **UI**: a "Timeline" tab beside the graph on `/hosts/:hostId`, sharing the active time window from the compact time control. Type filter chips, text search, and per-row links to the owning process node in the graph (and the reverse: a node's "show in timeline" action). Row artifacts reuse the bridge pivots from the fleet-search story.
- Brush-to-zoom on the timeline (drag-select sets an absolute window) if cheap; otherwise defer to the flagging/export follow-up.

### Acceptance criteria

- The timeline shows exec, network, and DNS rows interleaved in time order for the active window, filterable by type and text, paginated server-side.
- A row links to its process node in the graph; a graph node links to its rows in the timeline.
- Graph and timeline always reflect the same time window.

### Tests

- Handler tests: interleaving order, type filters, pagination stability.
- UI tests: tab switching preserves window; filters drive server queries; cross-links carry the right anchors.

getvictor:
Completed. Both PRs merged: #610 (GET /api/hosts/{host_id}/timeline over the ClickHouse archive) and #612 (Timeline tab UI beside the graph, with graph<->timeline cross-links). Stayed open only because the PR titles referenced it without a "Closes" keyword.

Issue #582, opened 2026-07-02
Investigation UX: fleet-wide artifact search, table + bridge pivots (merges #418, #419)


getvictor:
Part of #577 · Wave 2 · Size **XL** · absorbs: #418 (process search endpoint), #419 (process search UI), both closed as merged into this story; depends on: #423 (honest truncation metadata)

### Goal

The global hunting surface, fleet-wide from the start: a server-side artifact search endpoint with filters, cursor pagination, and sort; a paginated filterable table UI; and the "bridge" pattern where every artifact on the host page pivots to a pre-scoped global search in one click. The equivalent of Defender Advanced Hunting, Falcon Event Search, and SentinelOne Deep Visibility.

### Why

The tree is for one scoped story; the table is for volume: across a 1h/6h window there are tens of thousands of processes no graph can show, and analysts need `path = grep`, `unsigned AND has_network`, `uid = 0 AND time in [t0,t1]` answered precisely. In every leading EDR, real querying lives in a global console while the host page only filters within the host, and the two are connected by pre-populated, entity-scoped handoffs (MDE "Go hunt"). The top scoping questions ("where else did this hash run", "who else talked to this IP") are cross-host by nature, which is why this ships fleet-wide with `host_id` as an optional filter rather than as a per-host endpoint later widened.

### Scope

- **server, endpoint** (absorbing #418, widened fleet-wide): a process search endpoint (final route in the openspec proposal) with query params for `host_id` (optional), `path` (prefix/contains), `hash`, `signing` state, `uid`/user, `time` range, `has_network`, `exit_reason`, plus `sort` and an opaque keyset `cursor`. Returns rows + `next_cursor` + `total_matched`. Filters compose (AND) and are applied in SQL, not in Go. New store query alongside `GetProcessTree`. Absorbs the retired benign-suppression residual: cap-aware server-side filtering and optional signed-Apple-platform exclusion live here, where filtering and the row budget are decided together.
- **server, artifact classes**: beyond processes, search network connections by remote IP and DNS queries by domain, same pagination contract. Indexes as needed for hash, remote address, and domain lookups.
- **UI, table** (absorbing #419): global `/search` route. Columns: time, host, process, parent, cmdline, user, signing, exit_reason. Filter chips mapping to endpoint params; cursor-based pagination; no client-side filtering of a giant payload.
[...]

Issue #577, opened 2026-07-02
Epic: investigation UX: alert-first navigation, host context, conviction evidence, fleet-wide pivots


getvictor:
## Epic: investigation UX: alert-first navigation, host context, conviction evidence, fleet-wide pivots

> **Closed 2026-07-05.** All wave 1-2 stories and #583/#585 shipped; the last remaining story, #584 (host isolation), was spun off as a standalone issue since it is a cross-context control-plane feature, not the UI-centric work the rest of the epic covered.

### Problem

Competitive research (2026-07-02) against CrowdStrike Falcon, SentinelOne Singularity, Microsoft Defender for Endpoint, Elastic Security, and Jamf Protect shows a consistent common denominator that our UI lacks. Practitioner sources agree on the workflow: analysts enter alert-first (not host-first), convict on command lines and code-signing evidence, hunt in a flat filterable event list beside the graph, and pivot fleet-wide on any artifact (hash, remote IP, domain, user, path) via a one-click bridge from the host context view into a global search console. Key references: Falcon triage methodology (https://medium.com/@InfoSecDion/falcon-triage-methodology-navigating-crowdstrikes-edr-for-incident-response-8f77470a22e5), SentinelOne Process Graph + Deep Visibility (https://www.sentinelone.com/blog/introducing-the-new-singularity-xdr-process-graph/), MDE device investigation + "Go hunt" bridge (https://learn.microsoft.com/en-us/defender-endpoint/investigate-machines), Elastic visual event analyzer + Session View (https://www.elastic.co/guide/en/security/current/visual-event-analyzer.html).

Where we stand today:

- The app lands on the Hosts list (`/` renders `HostList`); Alerts is the second tab. The dominant analyst entry path is the alert queue.
- The host detail page (`/hosts/:hostId`) titles itself with the raw host id. Hostname, OS version, agent version, last seen, and source IP exist server-side but never render on this page (and the identity fields go stale until #437 lands).
- Command line and signing evidence are buried in the click-through detail panel. `team_id` and the platform flag are stored but not rendered; nothing is copyable in one click. Walking an ancestry chain costs one click plus panel scan per node.
- No fleet-wide search surface exists, and no artifact on the page pivots anywhere.
- Time selection is four fixed relative presets with no absolute range.
- Kill process is the only response action; the `isolate` command type is reserved server-side but unimplemented and has no UI.

### Relationship to existing issues

[...]

PR #612, opened 2026-07-04
feat(ui): host event timeline tab beside the process graph (#583, PR 2/2)

getvictor:
## What

Completes #583 (epic #577): the UI half of the host event timeline. The host page now offers a flat, filterable event stream beside the process graph, over the same time window, with cross-links between the two views. PR 1 (#610) shipped the `GET /api/hosts/{host_id}/timeline` endpoint this consumes.

- The host page gains a **Graph / Timeline** view selector; the active view lives in the URL (`?view=`), so a switch is bookmarkable and the shared time window + alert anchor survive it. Graph stays the default, so every existing host-page link still lands on the graph.
- `HostTimeline` renders the host's exec/network/DNS events for the active window as a flat newest-first table (time, type, process, per-type detail), filterable by event-type chips and a text box, with keyset "Load more" paging and a match count, driven by `getHostTimeline` over the shared `useCursorList` + `SearchResultsFrame`.
- **Cross-links both ways:** a timeline row opens its process in the graph (`?view=graph&pid=&at=`, resolved to the node whose lifetime brackets the anchor via `findNodeByPidAtTime`); a process node's detail panel gains "Show in timeline" (`?view=timeline&pid=`) which emphasizes that process's rows. Connection/DNS rows carry the same fleet-wide "search" pivot as the detail panel.
- `ProcessTreeView` guards its graph-only effects (tree fetch + d3 render) when the timeline is active, so the timeline view does no graph work; the shared `HostHeader`, time control, and activity histogram render in both views.

## Bugs found + fixed during manual QA

- **Blank-page crash on an empty result.** A zero-match page marshals its empty Go slice as JSON `null`, so the list consumer read `null.length`. `searchProcesses`/`searchEvents`/`getHostTimeline` now normalize a null `rows`/`events` to `[]`; this also hardens the merged #582 search paths, which had the same latent crash on an empty result (never hit live because those QAs returned non-empty).
- **Text filter dropped characters.** Its input was bound directly to the URL param, so each keystroke's `setSearchParams` re-render reset the native input (typing "python3" left only "3"). It now uses local input state debounced into the URL, so fast typing is never reset and a burst of keystrokes issues one query.

## Tests

`HostTimeline` (interleaved rows, type-chip filter, debounced text filter with no dropped chars, row->graph link target, emphasized pid, connection pivot, load-more, empty state); `ProcessTreeView` (view switch shares the window, `?pid=&at=` selects the owning node); `ProcessDetail` (show-in-timeline target); `getHostTimeline` URL composition + null-events normalization.

Full vitest 587, eslint `--max-warnings=0` clean, tsc clean, vite build clean, spectrace 100% (5 new scenario markers), `openspec validate --all --strict` 59/59.

## Manual QA

Dev server, host `AAAA1059`: the Timeline lists the 7 exec events with process links and command lines; text filter `python3` narrows to 2 with the full value retained (dropped-char bug fixed); a row opens the graph at `?view=graph&pid=5001&at=...` with the window preserved (shared-window + row->graph pivot confirmed).

## Spec

New openspec change `host-timeline-ui`: `web-ui` ADDED "Host event timeline view" and "Graph and timeline cross-navigation".


getvictor:
@coderabbitai review

PR #613, opened 2026-07-04
feat(ui): inline MITRE technique tags on nodes, panel, and timeline rows (#585)

### LINKED ISSUE #585: Investigation UX: inline MITRE technique tags on nodes and event rows
Part of #577 · Wave 3 · Size **S** · depends on: timeline story for row tags (node tags have no dependency)

### Goal

Show MITRE ATT&CK technique tags inline where investigation happens: on alerted graph nodes and on timeline rows, not only on the alert breadcrumb.

### Why

Inline technique mapping on events/nodes is table stakes across MDE (bold technique rows in the device timeline), SentinelOne (tactic-colored indicators on graph nodes), and Falcon. We already store technique IDs per alert and render badges on the alert breadcrumb only, so the mapping is invisible during the actual walk of the tree or the event stream.

### Scope

- **UI**: technique badge(s) on alerted nodes' hover tooltip and detail-panel alert entries (data already fetched); technique tags on timeline rows that belong to an alerted process once the timeline lands. Badges link to the existing `/rules/:ruleId` doc page like the breadcrumb badges do.

### Acceptance criteria

- An alerted node's tooltip and panel show its technique tags without navigating to the alert.
- Timeline rows for alerted processes carry the technique tag.
- Tags link to the rule documentation page.

### Tests

- UI component tests for tag rendering in tooltip, panel, and timeline row states.

getvictor:
## What

Closes #585 (epic #577, wave 3). Technique ids are stored per alert but rendered only on the alert breadcrumb, so the MITRE mapping is invisible during the actual investigation. This surfaces the tags inline where the analyst already is: the process graph, the detail panel, and the host timeline.

- A shared `TechniqueTags` component renders a technique-id list either as display-only badges (the hover tooltip, where a link in a transient card is unreachable) or as `/rules/:ruleId` link badges (panel + timeline).
- **Process graph**: an alerted node's hover tooltip lists its techniques. `ProcessTreeView` builds a `techniquesByNodeId` map (process DB id -> deduped technique ids) from the alerts it already fetches and threads it through `buildNodeTooltip`.
- **Detail panel**: each alert entry shows its techniques as badges linking to the rule doc page (data already fetched).
- **Host timeline**: a row whose event triggered an alert shows that alert's techniques, linked to the rule. Keyed by event id (the alert's triggering `event_ids` via `getAlertDetail`), not the process, because a timeline event carries the OS pid, not the alert's process DB id, and the timeline view does not load the tree that would map one to the other. This tags the exact event that fired the technique (the MDE pattern).

No server, agent, or wire change: every technique id is already on the alert payload.

## Tests

`TechniqueTags` (link vs display-only vs empty); `node-tooltip` carries techniques; `ProcessTreeView` fetches open+acked alerts and builds the map without crashing on techniques; `ProcessDetail` alert entry links each technique to the rule page; `HostTimeline` tags only the triggering-event row.

Full vitest 597, eslint `--max-warnings=0` clean, tsc clean, vite build clean, spectrace 100% (3 new scenario markers), `openspec validate --all --strict` 62/62.

## Manual QA

Dev server, seeded alert (rule `suspicious_exec`, techniques T1059.004 + T1105) on a host with a rendered tree:

- Hovering the alerted node shows the two techniques in the tooltip (the one render path unit tests cannot drive through jsdom's d3).
- The detail panel's alert entry shows the same two techniques as badges linking to `/rules/suspicious_exec` (verified the hrefs in the DOM).

The timeline-row tag is unit-verified and its event_id/techniques data path is API-confirmed; a live check was not possible because no seeded host has both an alert and its triggering events in the ClickHouse archive.

## Spec

New openspec change `inline-mitre-tags`: `web-ui` ADDED "Inline MITRE technique tags".

## Summary by CodeRabbit

* **New Features**
  * Added inline MITRE ATT&CK technique tags across investigation views.
  * Technique IDs now appear in process hover tooltips, process detail alerts, and host timeline rows.
  * When available, technique tags can link directly to the related rule documentation page.

* **Bug Fixes**
  * Timeline technique tags are shown only for the event that triggered the alert.
  * Empty or missing technique data no longer displays tags.


getvictor:
@coderabbitai review

PR #618, opened 2026-07-05
refactor: consolidation-pass cognitive-complexity paydown + deferred fixes [no-behavior-change] [no-docs-change]

getvictor:
## What & why

`consolidation-pass` maintenance paydown of the grandfathered cognitive-complexity stock (the monthly mandate in `docs/maintenance/tasks/consolidation-pass.md`), plus two small deferred fixes surfaced during the 2026-07-05 maintenance rerun. Everything is behavior-preserving; no observable behavior changes.

| Commit | Change |
|---|---|
| `config` | Centralize the required `EDR_CLICKHOUSE_DSN` check in `config.Load` via `requireStr` (both binaries checked it independently after Load). Observable boot behavior unchanged. |
| `ui/cmdline.ts` | Fix the one open SonarCloud issue (S4624: nested template literal in `quoteArg`) by extracting the inner expression. |
| `demo-seed/seed.go` | Extract `maybeRefreshExisting`, `newestReplayedTimestampNs`, `applyTimestampSlide`. |
| `spectrace/report.go` | Extract `forEachInFlightChangeDir` + `scanRemovedRequirementsFile` (also removes the lexical dup between the two parse functions). |
| `scale/runner.go` | Extract `makeHostStates`, `hostState.recordResult`, `foldHostLatencies`, `evaluatePassGate`, `validate{Scenario,Backlog}Options`. |
| `identity/bootstrap.go` | Extract the OIDC/SSO/break-glass/seed wiring closures; `New`'s wiring order + dependency set unchanged. |
| `appcontrol_handler.go` | Dedup the identical decode->validate->actor prelude across 7 mutation handlers into `readAppControlBody` / `actorOrInternalError` / `decodeAppControlBody{,IfPresent}`. |
| `ProcessTree.tsx` | Move the pure functions + the three `useMemo` bodies + `renderTree`'s branch callbacks into a new pure `ProcessTree.helpers.ts`. No hook order / effect deps / D3 join / state shape changed. |

Every Go refactor is extract-only (verbatim moves / closure-to-parameter): SQL literals, error strings/wrapping, evaluation order, status codes, response bodies, authz gating, and wiring order are all unchanged. The UI refactor is pure-helper extraction with no rendering-logic change.

## Checklist

- [x] **OpenSpec / behavior**: no observable behavior change (pure extract-refactors + the one config-validation-site move that preserves boot behavior), asserted via `[no-behavior-change]`.
[...]

PR #620, opened 2026-07-05
refactor: consolidate shared HTTP + HMAC-token primitives [no-docs-change]

getvictor:
Monthly consolidation pass (`docs/maintenance/tasks/consolidation-pass.md`). Baseline: dup 0.6% (23 blocks / 618 lines), aggregate cognitive complexity 6706 over 65712 ncloc. Two coherent clusters of genuine semantic duplication, both **behavior byte-identical** (no wire change, no spec/docs delta).

## Cluster 1: shared HTTP handler primitives → `server/httpserver`

- **`ReadCappedBody` / `DecodeCappedJSON`** replace 5 hand-rolled copies of the "read `limit+1`, reject when `len > limit`" body-cap pattern (the exact truncate-vs-reject invariant CodeRabbit flagged on #618) across two contexts: operator `appcontrol`/`detectionconfig` + identity `useradmin`/`saadmin`/`ssoadmin`. Each caller maps the returned outcome to its own typed error code, so the invariant lives and is tested in one place.
- **`WriteJSONError`** centralizes the identity `{"error": code}` envelope behind the 3 byte-identical `writeErr` wrappers (48 call sites unchanged).

Call sites use guard-ifs rather than a `switch` to avoid an `exhaustive`-lint dead case for the outcome a given caller cannot produce.

## Cluster 2: cross-context token envelope → `internal/hmactoken`

The self-validating `v1.<payload>.<mac>` HMAC-SHA256 envelope (`split`/`mac`/`Seal` + the strict-base64 decode, MAC-width check, and constant-time `hmac.Equal` verify core) was **mirrored** in `endpoint/internal/signedtoken` (host bearer token) and `identity/internal/satoken` (service-account access token) because ADR-0004 forbids those contexts importing each other. The `satoken` doc even lamented the format was "mirrored here rather than shared."

Extracted the security-critical core to a new shared `internal/hmactoken` (shared **through `internal/`**, not sideways). Each context package keeps its own `Claims` shape, its own post-verify claim checks (kid always; audience only `satoken`; expiry always), and its own error sentinels, mapping `hmactoken.ErrBadSignature`/`ErrMalformed` back to its own so callers' `errors.Is` checks are unchanged. Exported API of both packages is unchanged.

**This realizes ADR-0013's stated intent** ("generalize the #454 host-token machinery rather than introduce a second mechanism" / "reusing the existing signer") in an ADR-0004-compliant way, resolving the exact tension the old `satoken` doc recorded. `arch-go.yml` gains a reviewed `**.hmactoken` allowlist entry for both contexts (same mechanism as the existing `**.httpserver` / `**.secretseal` shared entries).

## Verification

- `go build ./...`; `go vet -tags integration ./server/... ./internal/...`; `task lint:arch` (arch-go) — all green
- custom golangci-lint — **0 issues**; gofmt clean
- Both token packages' existing PBT round-trip + tamper + malformed-table suites pass **unchanged** (the behavior guard); new `hmactoken` tests cover round-trip / tamper (incl. strict-decode malleability) / malformed / wrong-key / validation
- Coverage: `hmactoken` **97.5%**, `signedtoken` 92.6%, `satoken` 87.5%, all above the 80% new-code gate

`[no-docs-change]`: the Docs gate fires on the touched `*handler.go` files, but every per-site status/code/message is byte-identical, so nothing a user sees changed.

Deferred to a future pass (per the task's refuse-compounded-scope rule): `ProcessTree.tsx` cog-253 (top hotspot, warrants its own PR) and the operator two-field error-writer shape.

## Summary by CodeRabbit

* **New Features**
  * Added consistent, shared handling for signed tokens across services, improving interoperability and token validation reliability.
  * Added standardized capped request-body and JSON decoding helpers for more predictable API behavior.

* **Bug Fixes**
  * Improved rejection of malformed or tampered tokens.
  * Made large or invalid request bodies return clearer, consistent error responses without truncation.


getvictor:
@coderabbitai review

getvictor:
Dispositioning the 4 gemini-code-assist findings on the new `httpserver` body helpers as won't-fix (resolving the threads):

- **`r.Body == nil` guard (body.go:32) + its test:** for server requests net/http guarantees a non-nil Body, and every caller of `ReadCappedBody`/`DecodeCappedJSON` is a server HTTP handler. The pre-refactor code used the same `io.ReadAll(io.LimitReader(r.Body, ...))` with no nil guard, so a nil Body is neither reachable nor a regression. Per the repo's no-speculative-edge-cases rule (CLAUDE.md), a guard no caller can trigger is dead code.
- **panic on `*json.InvalidUnmarshalError` (body.go:52) + its test:** every caller passes a non-nil pointer (`&req` / `&v`). A non-pointer `dst` is a programmer error already surfaced as `BodyInvalidJSON`, not a crash; adding a panic path for an input no caller produces is the same dead-defensive-code pattern.

Behavior is byte-identical to the pre-consolidation handlers. Merging on green.

